PR #8, opened 2020-08-27
almost solution

Andrii256:
https://andrii256.github.io/js_employees_table_DOM/

я не знаю чого в мене toLocaleString не працює

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,247 @@
 'use strict';
 
-// write code here
+// start everything
+initSorting();
+initSelection();
+initEdition();
+initFormAddingAndWorking();
+
+// init sorting functionality
+function initSorting() {
+  // init necessary helpfull data
+  const tbody = document.querySelector('tbody');
+  const ascOrders = {
+    name: true,
+    position: true,
+    office: true,
+    age: true,
+    salary: true,
+  };
+
+  // init heplhull functions
+  function appendListToTbody(list) {
+    list.forEach(tr => {
+      tbody.append(tr);
+    });
+  }
+
+  function setOtherToDefault(exception) {
+    for (const key in ascOrders) {
+      if (key !== exception) {
+        ascOrders[key] = true;
+      }
+    }
+  }
+
+  // init sorting sunctions
+  function sortNumbers(order, columnName) {
+    const list = [...tbody.querySelectorAll('tr')];
+
+    if (ascOrders[columnName]) {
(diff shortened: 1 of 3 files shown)


dan3si (approved):

src/scripts/main.js
```diff
@@ -1,3 +1,248 @@
...
+  }
+
+  // init events
+  document.querySelector('tr').addEventListener('click', (event) => {
+    switch (event.target.innerText) {
+      case 'Name': sortStrings(0, 'name'); break;
+      case 'Position': sortStrings(1, 'position'); break;
+      case 'Office': sortStrings(2, 'office'); break;
+      case 'Age': sortNumbers(3, 'age'); break;
+      case 'Salary': sortSalaries(4, 'salary'); break;
```
irenhh, line 96: usually cases are written on the separate lines

src/scripts/main.js
```diff
@@ -1,3 +1,248 @@
...
+
+  function stopEditing() {
+    currentEdited.parentElement.innerText = (currentEdited.value)
+      ? currentEdited.value
+      : previousTextValue;
+    currentEdited.remove();
+    currentEdited = null;
+  }
+
+  tbody.addEventListener('dblclick', (event) => {
```
irenhh, line 140: destructure event

src/scripts/main.js
```diff
@@ -1,3 +1,248 @@
...
+
+  document.addEventListener('keydown', (event) => {
+    if (currentEdited && event.code === 'Enter') {
+      stopEditing();
+    }
+  });
+}
+
+// init form functionality
+function initFormAddingAndWorking() {
```
irenhh, line 181: initForm?

irenhh (commented):
image
don't allow letters after editing

src/scripts/main.js
Andrii256: here are more readable

src/scripts/main.js
Andrii256: why?

src/scripts/main.js
Andrii256: no, initForm is tooo general

Andrii256:
about letters after editing. Now it needs to change structure of functions. But these were not in task, so it is unnecessary (I can did this triffle, but it will get near 2 hour, and it will absolutely helpless experience. Our task is to became proffesionals and get job, so I want began react already. And it was not in technical task, so there are everything is fair)

PR #34, opened 2021-01-06
First version was completed

Serhii-Alieksieievych:
[DEMO LINK](https://Serhii-Alieksieievych.github.io/js_employees_table_DOM/)


aionova (approved):
Good job!
But your sort function doesn't work for new fields I add
https://prnt.sc/wz6w61

PR #36, opened 2021-01-09
Employee table

tamara-ostapets:
- [DEMO LINK](https://tamara-ostapets.github.io/js_employees_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -161,7 +161,11 @@ button.addEventListener('click', (e) => {
   e.preventDefault();
 
   const employeeForm = button.parentNode;
-  const { name : firstName, position, office, age, salary } = employeeForm.elements;
+  const { name: firstName,
+    position,
+    office,
+    age,
+    salary } = employeeForm.elements;
 
   const newEmployee = {
     firstName,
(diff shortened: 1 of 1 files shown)


khrulenko (approved):

src/scripts/main.js
```diff
@@ -1,3 +1,294 @@
...
+const header = document.querySelector('thead');
+const categories = header.querySelector('tr');
+const tbody = [...table.tBodies][0];
+const rows = tbody.rows;
+
+// sorting
+categories.addEventListener('click', (e) => {
+  const sorted = e.target.classList.contains('ascending')
+    ? sortDesc(e.target, rows, e.target.cellIndex)
+    : sortAsc(e.target, rows, e.target.cellIndex);
```
aliceblunt8, line 13: these 2 functions can be combined into one

src/scripts/main.js
```diff
@@ -1,3 +1,294 @@
...
+
+        return aConverted - bConverted;
+      });
+
+      break;
+  }
+
+  [...categories.cells].forEach(cell => cell.classList.remove('ascending'));
+
+  target.classList.add('ascending');
```
aliceblunt8, line 54: instead of adding and removing classes to DOM elements, you can create a variable setAscending, which will store a boolean value, which you can set to opposite on every click and sort table according whether it is true or false

aliceblunt8 (approved):
@tamara-ostapets nice solution, check the comments
also it would be nice to refactor a code a little bit and move each action to the separate function, so it'll be easier to understand what's going on
like this

createForm();
addTableSorting();
addRowSelecting();
addDataAppending();
addTableEditing();

PR #77, opened 2021-06-04
Employees table

ivan-major:


src/scripts/main.js
```diff
@@ -1,3 +1,243 @@
...
+  const cell = e.target;
+
+  const cellText = cell.textContent;
+
+  cell.textContent = '';
+
+  const cellInput = document.createElement('input');
+
+  cellInput.setAttribute('class', 'cell-input');
+  // cellInput.setAttribute('value', cellText)
```
vkryvytskyy, line 226: remove comments

vkryvytskyy (changes requested):
https://user-images.githubusercontent.com/38065883/120813672-c81af000-c556-11eb-9095-c8b6324caa2a.mov

The input value should contain replaced by input text.
Only one cell can be edited at the time.
On blur save changes to table cell. Remove input and set new text.
On 'Enter' keypress, save changes to the table cell. Remove input and set new text in the table cell.
If an input is empty on submitting return initial value.

Вот эти пункты не выполнены

vkryvytskyy (approved):
image
Такого не должно быть

PR #121, opened 2022-01-21
That was tough, but interesting!

DimaBrushnivskyi:


danylo-onishchuk (changes requested):
форма не очищается после записи

нотификашки налазят друг на друга 

Screenshot 2022-01-24 at 11 30 46

src/scripts/main.js
```diff
@@ -1,3 +1,285 @@
...
+ageLabel.append(ageInput);
+salaryLabel.append(salaryInput);
+
+// Appention of the labels to the form
+form.append(nameLabel);
+form.append(positionLabel);
+form.append(selectLabel);
+form.append(ageLabel);
+form.append(salaryLabel);
+form.append(button);
```
aliceblunt8, line 84: another way to do this (in my humble opinion more declarative and neat) is to use insertAdjacentHTML method and add markup with all the necessary information embedded inside, but the you wrote is also valid
image

src/scripts/main.js
```diff
@@ -1,3 +1,285 @@
...
+const buttonSave = document.querySelector('button');
+const tBody = document.querySelector('tbody');
+
+buttonSave.addEventListener('click', function addNewEmployee(action) {
+  action.preventDefault();
+
+  const target = action.target.closest('button');
+
+  // Create new employee from the Form inputs
+  class NewEmployee {
```
aliceblunt8, line 96: nice 👍

aliceblunt8 (approved):
good job!
also to make your code look more structured you can encapsulate certain parts of logic in separate functions and call them in the right order at the beginning of the script 

```
createForm();
addTableSorting();
addRowSelecting();
addDataAppending();
addTableEditing();
```

PR #137, opened 2022-03-10
v1

KristinaTar:


vitaliy-martynyuk (changes requested):
everything seems good, only a few fixes left. it is possible to specify numbers in `Name` and `Position` fields. also, all fields are not cleared after adding new employee to the table
Screen Shot 2022-03-11 at 10 43 22 AM

vitaliy-martynyuk (approved):

PR #156, opened 2022-06-17
ass burned

Vovahunter:
- [DEMO LINK](https://vovahunter.github.io/js_employees_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,188 @@
 'use strict';
 
-// write code here
+const table = document.querySelector('table');
+const tabBody = document.querySelector('tbody');
+const body = document.querySelector('body');
+
+table.addEventListener('click', eve => {
+  const item = eve.target.closest('th');
+  const idx = item.cellIndex;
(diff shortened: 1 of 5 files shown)


aliceblunt8 (changes requested):
- sorting should work in two directions
Screenshot 2022-06-18 at 15 17 04
- this task isn't done
> Implement editing of table cells by double-clicking on it.

src/scripts/main.js
```diff
@@ -1,3 +1,189 @@
...
+
+    return false;
+  }
+
+  return true;
+}
+
+tHead.addEventListener('click', (e) => {
+  const index = e.target.cellIndex;
+  const isSorted = e.target.matches('.sorted');
```
SanyaBratashchuk, line 87: It is better to use CSS classes only for styling purposes. Use JS variable to toggle your sort order

SanyaBratashchuk (approved):
Good job! Check comment and If you have time and desire, you can try implementing validation on blur

PR #168, opened 2022-07-07
created employees table

silchyn:
[DEMO](https://silchyn.github.io/js_employees_table_DOM/)


VETALEK (changes requested):
GJ 🔥 
need to add validation on editing records

image

PR #173, opened 2022-07-12
employees-table-but-not-all

TanyaHryn:


Bohdan-Maksymiuk (changes requested):
> Implement table sorting by clicking on the title (**in two directions**).
> Show notification if form data is invalid (use notification from the previous tasks).

now data is not validated and I can submit empty form
Screenshot 2022-07-15 at 07 31 20

PR #181, opened 2022-07-22
solution (js_employees_table_DOM)

nosovandriy:
[DEMO LINK](https://nosovandriy.github.io/js_employees_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,259 @@
 'use strict';
 
-// write code here
+const thead = document.querySelector('thead');
+const tbody = document.querySelector('tbody');
+let sortTargetInnerText = '';
+let sortDirection = 1;
+
+// 1. sorting
+thead.addEventListener('click', (eventClick) => {
+  const cellIndex = eventClick.target.cellIndex;
+  const trArray = [...tbody.children];
+
+  if (sortTargetInnerText !== eventClick.target.innerText) {
+    sortDirection = 1;
+  };
+
+  eventClick.target.dataset.clicked = eventClick.target.innerText;
+
+  const arraySorted = trArray.sort((a, b) => {
+    const first = a.children[cellIndex].innerText.replace(/[$,]/g, '');
+    const second = b.children[cellIndex].innerText.replace(/[$,]/g, '');
+
+    if (isNaN(first)) {
+      return sortDirection * first.localeCompare(second);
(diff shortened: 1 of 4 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,259 @@
 'use strict';
 
-// write code here
+const thead = document.querySelector('thead');
+const tbody = document.querySelector('tbody');
+let sortTargetInnerText = '';
+let sortDirection = 1;
```
ArtemTopolskyi, line 6: what means 1? lets avoid using magic numbers system and use understandable constants

src/scripts/main.js
```diff
@@ -1,3 +1,259 @@
...
+    ? document.createElement('select')
+    : document.createElement('input');
+
+  newInput.className = 'cell-input';
+
+  newInput.name = `${theadInnerText}`;
+  newInput.dataset.qa = `${theadInnerText}`;
+  newInput.value = eventClick.target.innerText;
+
+  switch (cellIndex) {
```
ArtemTopolskyi, line 183: again magic numbers, lets create understandable constants

src/scripts/main.js
```diff
@@ -1,3 +1,259 @@
...
+  newInput.focus();
+
+  newInput.addEventListener('keydown', (eventKey) => {
+    if (eventKey.key === 'Enter') {
+      newInput.blur();
+    };
+  });
+
+  newInput.addEventListener('blur', () => {
+    switch (true) {
```
ArtemTopolskyi, line 229: don't use swithc (true) construction, use if-else instead. You don't need switch construction if you write switch(true)

ArtemTopolskyi (changes requested):

src/scripts/main.js
nosovandriy: This is a change of sort direction.
When multiplying by 1, it is asc sorted
When multiplying by -1, it is desc sorted

src/scripts/main.js
nosovandriy: What does mean magic numbers ? These are indexes cells

src/scripts/main.js
nosovandriy: Ok, corrected it

ArturErmolenko (approved):

PR #184, opened 2022-07-30
solve employeesTableDOM problem

tsybulko-natalya:
https://tsybulko-natalya.github.io/js_employees_table_DOM/


marina-tilniak (changes requested):
image
 let's implement editing

PR #201, opened 2022-09-07
Develop

TetianaKoval:
[DEMO LINK](https://TetianaKoval.github.io/js_employees_table_DOM/)


AlisaVlasova (changes requested):
image
If an input is empty on submitting return initial value.

On 'Enter' keypress, save changes to the table cell. Remove input and set new text in the table cell.

Virass (approved):

PR #204, opened 2022-09-09
employees_table

Blenemy:
- [DEMO LINK](https://Blenemy.github.io/js_employees_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,201 @@
 'use strict';
 
-// write code here
+const body = document.querySelector('body');
+const tbody = document.querySelector('tbody');
+const error = document.createElement('div');
+const savingObject = {
+  state: true,
+};
+
+error.classList.add('error');
+error.classList.add('notification');
+error.setAttribute('data-qa', 'notification');
+
+const success = document.createElement('div');
+
+success.classList.add('success');
+success.classList.add('notification');
+success.innerHTML = 'You were successfully added to the table';
+success.setAttribute('data-qa', 'notification');
+
+function convertFromForm(number) {
+  return '$' + parseInt(number).toLocaleString('de-DE').split('.').join(',');
+}
+
+function converter(number) {
+  return number.slice(1).split(',').join('.');
+}
+
+body.insertAdjacentHTML('beforeend', `
+<form class="new-employee-form" action="">
+  <label  for="name">Name:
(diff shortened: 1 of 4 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,201 @@
...
+    .sort((a, b) => a.children[position]
+      .innerHTML > b.children[position].innerHTML ? 1 : -1);
+
+  if (position === 4) {
+    sorted = [...trs]
+      .sort((a, b) => converter(a.children[position]
+        .innerHTML) - converter(b.children[position].innerHTML));
+  }
+
+  if (order === false) {
```
Virass, line 73:
```suggestion
  if (!order) {
```

src/scripts/main.js
```diff
@@ -1,3 +1,201 @@
...
+    sorted = [...trs]
+      .sort((a, b) => b.children[position]
+        .innerHTML > a.children[position].innerHTML ? 1 : -1);
+
+    if (position === 4) {
+      sorted = [...trs]
+        .sort((a, b) => converter(b.children[position]
+          .innerHTML) - converter(a.children[position].innerHTML));
+    }
+  }
```
Virass, line 83: Code dublicated. Fix it.

```suggestion
  if (position === 4) {
    sorted = [...trs]
      .sort((a, b) => converter(a.children[position]
        .innerHTML) - converter(b.children[position].innerHTML));
  }

  if (order === false) {
    sorted = [...trs]
      .sort((a, b) => b.children[position]
        .innerHTML > a.children[position].innerHTML ? 1 : -1);

    if (position === 4) {
      sorted = [...trs]
        .sort((a, b) => converter(b.children[position]
          .innerHTML) - converter(a.children[position].innerHTML));
    }
  }
```

Virass (changes requested):
What specific input is incorrect? And fix the style for an error message. Example add padding

image

src/scripts/main.js
Blenemy: it doesnt work this way because i need order to be false, if (!order)  means that it doesnt exist

src/scripts/main.js
Blenemy: i dont understand, the code is not dublicated, if false AND if position === 4 i need the table to be sorted in desc order

ArturErmolenko (approved):

PR #212, opened 2022-09-23
Develop

grafit-off:
- [DEMO LINK](https://grafit-off.github.io/js_employees_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -274,7 +274,7 @@ form.addEventListener('submit', (e) => {
     name: form.elements.name.value,
     position: form.elements.position.value,
     office: form.elements.office.value,
-    age: form.elements.age.value,
+    age: Number(form.elements.age.value),
     salary: formatSalaryToNumber(form.elements.salary.value),
   };
 
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,284 @@
 'use strict';
 
-// write code here
+function formatSalaryToNumber(salary) {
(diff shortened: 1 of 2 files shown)


TheVovchik (changes requested):
1) I have an error message that don't help me at all
2) If `Name` value has less than 4 **letters**, show error notification. I type 4 numbers and validation don't stop me;
3) I think after success we need clear up form (because I can add infinit numbers of same workers, what not looks great)
4) - When users click on a new title, always sort in ASC order. - doesn't work

grafit-off:
> 1. I have an error message that don't help me at all
> 2. If `Name` value has less than 4 **letters**, show error notification. I type 4 numbers and validation don't stop me;
> 3. I think after success we need clear up form (because I can add infinit numbers of same workers, what not looks great)
> 4. * When users click on a new title, always sort in ASC order. - doesn't work

**_Thank you for your feedback! Already fixed!_**

src/scripts/main.js
```diff
@@ -1,3 +1,312 @@
...
+      const sortMultiplier = asc ? 1 : -1;
+
+      return prev[field].localeCompare(curr[field]) * sortMultiplier;
+    }
+
+    if (asc) {
+      return prev[field] - curr[field];
+    }
+
+    return curr[field] - prev[field];
```
naziamoff, line 35:
```suggestion
    return asc 
      ?  prev[field] - curr[field]
      :  curr[field] - prev[field]
```

PR #217, opened 2022-09-29
solution

vitalii-babiienko:
[DEMO LINK](https://vitalii-babiienko.github.io/js_employees_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,161 @@
 'use strict';
 
-// write code here
+const body = document.querySelector('body');
+const tableHead = document.querySelector('thead');
+const tableBody = document.querySelector('tbody');
+const rows = [...tableBody.rows];
+let orderASC = true;
+let lastIndex;
+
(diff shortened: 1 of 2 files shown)


AlisaVlasova (changes requested):
image
sorting doesn`t work properly, on the screen there should be sorted by salary
also added salary should be same format as all

image
this part isn`t implemented

src/scripts/main.js
```diff
@@ -1,3 +1,317 @@
...
+    return;
+  }
+
+  cellForEditing.contenteditable = true;
+  initialCellValue = cellForEditing.textContent;
+
+  switch (cellForEditing.cellIndex) {
+    case 3:
+    case 4:
+      cellForEditing.innerHTML = `
```
TarasHoliuk, line 209: image
add `value` attribut with cell text

TarasHoliuk (changes requested):
Good job! but small fix is needed

SanyaBratashchuk (approved):

PR #226, opened 2022-10-16
js_employees_table_DOM

dima-korovai:
[DEMO LINK](https://dima-korovai.github.io/js_employees_table_DOM/)


polosanya (changes requested):
It looks a lit bit strange. Think about adding z-index to your notification
image
Your page is crushing if name\position is a number
image
If user clicks on the title and then on another title, sorting is in DESC order, not ASC
image

dima-korovai:
image

About second point "crushing if name\position is a number" I don't understand. When I introduce the same dates it doesn't allow me to send the form while "Name" or "Position" have less than 4 digits, even if is a number.

src/scripts/main.js
```diff
@@ -1,3 +1,323 @@
...
+
+        case 'number':
+          return rowB.cells[colNum].textContent
+          - rowA.cells[colNum].textContent;
+        case 'salary':
+          return toNumber(rowB.cells[colNum].textContent)
+          - toNumber(rowA.cells[colNum].textContent);
+      }
+    };
+  }
```
loralevitska, line 68: this code looks like duplicate. Try to optimize here. As I understood from your code, it will be enough to use 1 function and pass parameters to it in different order

```suggestion
  if (order === 'ASC') {
    compare = function(rowA, rowB)
  } else {
    compare = function(rowB, rowA)
  }
```

src/scripts/main.js
```diff
@@ -1,3 +1,323 @@
...
+
+const button = body.querySelector('button');
+
+const inputName = document.querySelector('[data-qa ="name"]');
+const inputPosition = document.querySelector('[data-qa ="position"]');
+const inputOffice = document.querySelector('[data-qa ="office"]');
+const inputAge = document.querySelector('[data-qa ="age"]');
+const inputSalary = document.querySelector('[data-qa ="salary"]');
+
+const onlyWithLabels = [...form].filter(element =>
```
loralevitska, line 175: The filter() method creates a new array with all elements that pass the test specified in the passed function.
```suggestion
const onlyWithLabels = form.filter(element =>
```

src/scripts/main.js
```diff
@@ -1,3 +1,323 @@
...
+    .filter(elem => elem.dataset.qa === 'age')
+    .every(el => el.value > 17 && el.value < 91);
+
+  const correctInputSalaryLength = [...onlyWithLabels]
+    .filter(elem => elem.dataset.qa === 'salary')
+    .every(el => el.value.length > 0);
+
+  if (correctInputTextLength
+    && correctInputAgeLength
+    && correctInputSalaryLength) {
```
loralevitska, line 193: put long expressions in variables

src/scripts/main.js
```diff
@@ -1,3 +1,323 @@
...
+    return '$' + '0,' + input;
+  }
+
+  if (input.length === 2) {
+    return '$' + '0,0' + input;
+  }
+
+  if (input.length === 1) {
+    return '$' + '0,00' + input;
+  }
```
loralevitska, line 272: switch case will be better

loralevitska (changes requested):
you demo looks good)
validate salary for positive integer 
image
and check comments to improve your code

src/scripts/main.js
dima-korovai: image

if I write form.filter  and not [...form].filter it gets me error

src/scripts/main.js
dima-korovai: I don't sure I changed it correctly. Can you tell me if now it's ok please?

src/scripts/main.js
```diff
@@ -1,3 +1,318 @@
...
+};
+
+const preventSelection = function(ev) {
+  ev.preventDefault();
+};
+
+table.addEventListener('click', handleListRows);
+table.addEventListener('mousedown', preventSelection);
+
+// adding form
```
liliia-livandovskaya, line 95: remove comments

src/scripts/main.js
```diff
@@ -1,3 +1,318 @@
...
+    case 2:
+      return '$' + '0,0' + input;
+    case 1:
+      return '$' + '0,00' + input;
+    default:
+      return '$' + input.replace(/\B(?=(\d{3})+(?!\d))/g, ',');
+  }
+}
+
+// editing of table cells
```
liliia-livandovskaya, line 272:
```suggestion

```

liliia-livandovskaya (approved):
GJ 👍  Pay attention to the comments

PR #232, opened 2022-10-29
Solution

OksanaShvets1:
Help please with dbclick and notification - I waiting for your help
 - [DEMO LINK](https://OksanaShvets1.github.io/js_employees_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,251 @@
 'use strict';
 
 // write code here
+document.addEventListener('DOMContentLoaded', function() {
+  const table = document.querySelector('table');
+  const headers = table.querySelectorAll('th');
+  const tableBody = table.querySelector('tbody');
+  const rows = tableBody.querySelectorAll('tr');
+
+  const directions = Array.from(headers).map(function(header) {
+    return '';
+  });
+
+  const transform = function(index, content) {
+    const type = headers[index].getAttribute('data-type');
+
+    switch (type) {
+      case 'number':
+        return parseFloat(content.slice(1, -1));
+      case 'string':
+      default:
+        return content;
+    }
+  };
+
+  const sortColumn = function(index) {
+    const direction = directions[index] || 'asc';
+    const multiplier = (direction === 'asc') ? 1 : -1;
+
+    const newRows = Array.from(rows);
+
+    newRows.sort(function(rowA, rowB) {
+      const cellA = rowA.querySelectorAll('td')[index].innerHTML;
+      const cellB = rowB.querySelectorAll('td')[index].innerHTML;
+
+      const a = transform(index, cellA);
+      const b = transform(index, cellB);
+
+      switch (true) {
+        case a > b: return 1 * multiplier;
+        case a < b: return -1 * multiplier;
+        case a === b: return 0;
+      }
+    });
+
+    [].forEach.call(rows, function(row) {
+      tableBody.removeChild(row);
+    });
+
+    directions[index] = direction === 'asc' ? 'desc' : 'asc';
+
+    newRows.forEach(function(newRow) {
+      tableBody.appendChild(newRow);
+    });
+  };
+
+  [].forEach.call(headers, function(header, index) {
+    header.addEventListener('click', function() {
+      sortColumn(index);
+    });
+  });
+});
+
+const newForm = document.createElement('form');
+
+newForm.className = 'new-employee-form';
+
+newForm.innerHTML = `<label>Name:
+<input
+name="name"
+type="text" data-qa="name" required minlength="4" id="name" value=""></label>
+<label>Position: <input
+name="position"
+type="text" data-qa="position" required id="position" value=""></label>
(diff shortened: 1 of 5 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,251 @@
...
+      sortColumn(index);
+    });
+  });
+});
+
+const newForm = document.createElement('form');
+
+newForm.className = 'new-employee-form';
+
+newForm.innerHTML = `<label>Name:
```
TarasHoliuk, line 68: please, keep the code style in this HTML string template the same as in the default HTML code

src/scripts/main.js
```diff
@@ -1,3 +1,251 @@
...
+    const newRows = Array.from(rows);
+
+    newRows.sort(function(rowA, rowB) {
+      const cellA = rowA.querySelectorAll('td')[index].innerHTML;
+      const cellB = rowB.querySelectorAll('td')[index].innerHTML;
+
+      const a = transform(index, cellA);
+      const b = transform(index, cellB);
+
+      switch (true) {
```
TarasHoliuk, line 39: wrong usage of `switch case` construction
use `if` in such cases

src/scripts/main.js
```diff
@@ -1,3 +1,251 @@
 'use strict';
 
 // write code here
+document.addEventListener('DOMContentLoaded', function() {
+  const table = document.querySelector('table');
+  const headers = table.querySelectorAll('th');
+  const tableBody = table.querySelector('tbody');
+  const rows = tableBody.querySelectorAll('tr');
+
+  const directions = Array.from(headers).map(function(header) {
```
TarasHoliuk, line 10: no need to save smth like an array of directions. Anyway, you sort rows based on one chosen column. So save just string 'asc' || 'desc' etc.

src/scripts/main.js
```diff
@@ -1,3 +1,251 @@
...
+</label>
+<label>Age: <input name="age"
+  type="number"
+  data-qa="age" required min="18" max="90" id="age" value=""></label>
+<label>Salary: <input name="salary"
+  type="number" data-qa="salary" required id="salary" value=""
+  ></label>
+<button>Save to table</button>`;
+
+const tableN = document.querySelector('table');
```
TarasHoliuk, line 92: you don't need to do it
just use `table` variable

src/scripts/main.js
```diff
@@ -1,3 +1,251 @@
...
+const tody = document.querySelector('body');
```
TarasHoliuk, line 93:
```suggestion
const body = document.querySelector('body');
```

src/scripts/main.js
```diff
@@ -1,3 +1,251 @@
...
+    notification.textContent
+      = `Name to short`;
+  }
+
+  if (age.validity.rangeUnderflow) {
+    notification.textContent = 'you have to be adult';
+  } else if (age.validity.rangeOverflow) {
+    notification.textContent = 'you have to be younger';
+  }
+  document.querySelector('table').appendChild(notification);
```
TarasHoliuk, line 203:
```suggestion
  table.appendChild(notification);
```

src/scripts/main.js
```diff
@@ -1,3 +1,251 @@
...
+}
+
+let area = null;
+const view = document.querySelector('table');
```
TarasHoliuk, line 207:
```suggestion
```

src/scripts/main.js
```diff
@@ -1,3 +1,251 @@
...
+
+view.addEventListener('dblclick',
```
TarasHoliuk, line 209:
```suggestion
table.addEventListener('dblclick',
```

src/scripts/main.js
```diff
@@ -1,3 +1,251 @@
...
+    if (event.key === 'Enter') {
+      this.blur();
+    }
+  };
+
+  area.onblur = function() {
+    editEnd(td);
+  };
+
+  td.replaceWith(area);
```
TarasHoliuk, line 244: don't do replace
append input inside `td` tag instead of text, but save text before it
initial value of input should be saved (initiall) text

src/scripts/main.js
```diff
@@ -1,3 +1,251 @@
...
+nameEmployee.addEventListener('input', function() {
+  if (nameEmployee.validity.valid) {
+    notification.textContent = '';
+    notification.className = 'notification error';
+  } else {
+    showError();
+  }
+});
+
+age.addEventListener('input', function() {
```
TarasHoliuk, line 166: I guess you should validate and show error on submit, not after each change

TarasHoliuk (changes requested):
I've added a few suggestions
If you need help with a task feel free to ask in fe_chat

AlisaVlasova (changes requested):
save to table doesn't work
if you have questions it`ll be better to ask them in fe_chat or QnA meeting, that would be faster)

AlisaVlasova (changes requested):
image
знову сортування не працює з новим айтемом таблички(

nazarmatsevych (approved):

PR #237, opened 2022-11-05
the " js_employees_table_DOM" is done

Anastasiya145:
https://anastasiya145.github.io/js_employees_table_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,163 @@
 'use strict';
+// работа с ячейками
 
-// write code here
+const body = document.querySelector('body');
+const tbody = document.querySelector('tbody');
+
+tbody.addEventListener('click', (ev) => {
+  const activeRow = ev.target.closest('tr');
+
+  [...tbody.rows].forEach((row) => row.classList.remove('active'));
+  activeRow.classList.add('active');
+});
+
+body.insertAdjacentHTML('beforeend', `
+  <form class="new-employee-form" action="#" method="get">
+    <label>
+      Name:
+      <input
+        type="text"
+        name="name"
+        data-qa="name"
+        required
+      >
+    </label>
+    <label>
+      Position:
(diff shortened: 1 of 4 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,163 @@
...
+      ${text}
+    </p>
+  </div>
+  `);
+
+  setTimeout(() =>
+    body.removeChild(document.querySelector('.notification')), 2000);
+};
+
+function strEdited(str) {
```
TarasHoliuk, line 91: Don't use name cuts
Also, if it is a function, start the name with a verb that explains what the function does. For example `editString`
But in your case, you can even specify: `capitalizeString`

src/scripts/main.js
```diff
@@ -1,3 +1,163 @@
...
+  return (str[0].toUpperCase() + str.slice(1));
+};
+
+function salaryEdited(salary) {
+  const newFormat = new Intl.NumberFormat('en-US');
```
TarasHoliuk, line 96: Taking into account that below you're using `format` method of this variable, I'd name it `numberFormatter`

src/scripts/main.js
```diff
@@ -1,3 +1,163 @@
...
+    tbody.insertAdjacentHTML('afterbegin', `
+      <tr>
+        <td>${strEdited(nameData)}</td>
+        <td>${strEdited(positionData)}</td>
+        <td>${officeData}</td>
+        <td>${ageData}</td>
+        <td>${salaryEdited(salaryData)}</td>
+      </tr>
+    `);
+
```
TarasHoliuk, line 131: Clear form after submitting
image

TarasHoliuk (changes requested):
GJ! I've added just a few suggestions

TarasHoliuk (approved):

PR #239, opened 2022-11-06
cod for employers table

RusNeday:


src/scripts/main.js
```diff
@@ -1,3 +1,276 @@
...
+  const data = new FormData(form);
+  const informNewPerson = Object.fromEntries(data.entries());
+
+  if (informNewPerson.name.length < 4
+    || informNewPerson.age < 18
+    || informNewPerson.age > 90) {
+    pushNotification(100, 10,
+      'ERROR',
+      'It is necessary to fill in all fields'
+      + 'of the form to correct!',
```
vlad-demchuk, line 65: better create a separate variable for this string and set what is the problem with filling(age)

src/scripts/main.js
```diff
@@ -1,3 +1,276 @@
...
+});
+
+const sortPersonInTable = thead.addEventListener('click', (e) => {
+  const headerName = e.target.innerHTML;
+
+  choiceColumnSort(e, 'Name', headerName);
+  choiceColumnSort(e, 'Position', headerName);
+  choiceColumnSort(e, 'Office', headerName);
+  choiceColumnSort(e, 'Age', headerName);
+  choiceColumnSort(e, 'Salary', headerName);
```
vlad-demchuk, line 97: you call this function 5 times. it's not about performance. try to refactor this. you can use some conditionals or switch

vlad-demchuk (changes requested):

src/scripts/main.js
vlad-demchuk: Check all such cases. And try to make it simpler.

vlad-demchuk: UPD
form should be reset after submit
I can't edit added rows
salary should be required

PR #240, opened 2022-11-06
solution

Moblat890:


qoa11a (changes requested):
image
I can add people with invalid data. Fix it.

polosanya (changes requested):
1. If user fixed a problem before the error message disappears, this happens on success
image
2. Sorting doesn't work properly if a new employee was added
employee

SanyaBratashchuk (approved):

PR #242, opened 2022-11-10
Employees table functionality was implemented

mykyta-krainik:
[DEMO LINK](https://mykyta-krainik.github.io/js_employees_table_DOM/)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 1. Replace `<your_account>` with your Github username in the link
-    - [DEMO LINK](https://<your_account>.github.io/js_employees_table_DOM/)
+    - [DEMO LINK](https://mykyta-krainik.github.io/js_employees_table_DOM/)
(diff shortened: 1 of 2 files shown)


liliia-livandovskaya (changes requested):
1. The fields in the form should be clear after the form submit 
Screenshot 2022-11-10 at 13 40 04

2. Add validation on the editing 
Screenshot 2022-11-10 at 13 40 26

Anastasiia-Svintsova (changes requested):
- You have strange validation for the office and there is an issue connected with it, look at the videos below
Animation
Animation

polosanya (changes requested):
1. Error message looks a bit strange. I think its z-index must be higher 
image
2. Is it a bug or a feature? 😅
employee-table
3. It happens if user fixes a problem before error message disappears
image

PR #245, opened 2022-11-11
solution

illia-kyselov:
- [DEMO LINK](https://illia-kyselov.github.io/js_employees_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,280 @@
 'use strict';
 
-// write code here
+document.body.style.alignItems = 'flex-start';
+
+const headList = document.querySelector('thead');
+const headName = [
+  ...document.querySelectorAll('thead th')].map(i => i.innerText);
+
+const body = document.querySelector('tbody');
+const rows = [...body.querySelectorAll('tr')];
+const cell = (row, n) => row.children[n].textContent.replace(/[$,]/g, '');
+
+headList.addEventListener('click', (e) => {
+  let sorted;
+  const target = e.target.closest('th').innerText;
+  const numOfColum = headName.indexOf(target);
+
+  if (numOfColum >= 3) {
+    sorted = rows.sort((a, b) =>
+      cell(a, numOfColum) - (cell(b, numOfColum)));
+  } else {
+    sorted = rows.sort((a, b) =>
+      cell(a, numOfColum).localeCompare(cell(b, numOfColum)));
+  };
+  sorted.map(i => body.appendChild(i));
+});
+
+headList.addEventListener('dblclick', (e) => {
+  let sorted;
+  const target = e.target.closest('th').innerText;
+  const numOfColum = headName.indexOf(target);
+
+  if (numOfColum >= 3) {
+    sorted = rows.sort((a, b) =>
+      cell(b, numOfColum) - (cell(a, numOfColum)));
+  } else {
+    sorted = rows.sort((a, b) =>
+      cell(b, numOfColum).localeCompare(cell(a, numOfColum)));
+  };
+
+  sorted.map(i => body.appendChild(i));
+});
+
+const form = document.createElement('form');
+
+form.classList.add('new-employee-form');
+document.body.append(form);
+
+const citiesOptions = [
+  'Tokyo',
+  'Singapore',
+  'London',
+  'New York',
+  'Edinburgh',
+  'San Francisco',
+];
+const formInputs = [
+  ['Name', 'text'],
+  ['Position', 'text'],
+  ['Age', 'number'],
+  ['Salary', 'number'],
+];
+
+function addInput(inputName, inputType) {
+  const label = document.createElement('label');
+  const input = document.createElement('input');
+
+  label.innerText = inputName + ': ';
+  input.type = inputType;
+  input.name = inputName.toLowerCase();
+  input.dataset.qa = inputName.toLowerCase();
+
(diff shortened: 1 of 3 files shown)


TarasHoliuk:
Newly added rows don't sort
image

TarasHoliuk:
image
The second point: 
image

If you need help with it fell free to ask in chat

src/scripts/main.js
```diff
@@ -1,3 +1,280 @@
...
+  }
+
+  function saveEditedField() {
+    if (!input.value) {
+      changeCellContent();
+
+      return;
+    }
+
+    if ((isAge || isSalary) && !input.value.match(/[0-9]/)) {
```
TarasHoliuk, line 223: You are doing validation in `checkValidation`, and don't run the current function if validation fails, no need to check all fields here again. So `saveEditedField` is redundant, you can run `changeCellContent` instead of it.

TarasHoliuk (changes requested):

src/scripts/main.js
illia-kyselov: I didn't understand how it is possible to do the same without calling the function saveEditedField

src/scripts/main.js
```diff
@@ -1,3 +1,280 @@
...
+
+  pushNotification('Great!',
+    'New employee successfully added!',
+    'success'
+  );
+
+  return true;
+};
+
+button.addEventListener('click', (e) => {
```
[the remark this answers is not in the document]
illia-kyselov: image

And I don't understand what the 2nd point means.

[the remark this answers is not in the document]
TarasHoliuk: Feel free to ask in the chat if you still have questions

src/scripts/main.js
```diff
@@ -1,3 +1,283 @@
...
+    <td>${data.office}</td>
+    <td>${data.age}</td>
+    <td>$${Number(data.salary).toLocaleString('en-US')}</td>
+  `;
+
+  body.append(tr);
+};
+
+const pushNotification = (title, description, type) => {
+  const messageConteiner = document.createElement('div');
```
TarasHoliuk, line 124:
```suggestion
  const messageContainer = document.createElement('div');
```

TarasHoliuk: typo

TarasHoliuk (changes requested):

src/scripts/main.js
illia-kyselov: I don’t understand how to make a check so that when the first letter is entered, the previous text of the cell is deleted, if I write in this place `initialValue` then only after removing the focus, the previous text is deleted, and the new one remains
Desktop Screenshot 2022 11 20 - 17 02 38 94
Desktop Screenshot 2022 11 20 - 17 06 08 46
Desktop Screenshot 2022 11 20 - 17 06 28 46

src/scripts/main.js
illia-kyselov: I didn't make it up how rewrite it for both cases, in two cases different types of data and their validation and sooo I don't know how make it

src/scripts/main.js
illia-kyselov: I write my questions there because mentors in fe_chat don't know my all code and it will be very difficult explain to them my problems easily

qoa11a (changes requested):
https://www.loom.com/share/eaae8ba6d2f14f59b394f6786b00c167
I can't edit city for some reason. Fix it please.
Also when I edit salary one dollar sign is added always.

illia-kyselov:
> 


the cell about cities can only be changed to cities from the available list, bug with dollar was fixed

src/scripts/main.js
```diff
@@ -1,3 +1,301 @@
...
+    sortOrder = 'DESC';
+    sortHeader = null;
+  }
+
+  const forSortedTable = [...body.children].sort((a, b) => {
+    const cellA = a.cells[toSort].innerText;
+    const cellB = b.cells[toSort].innerText;
+
+    if (sortOrder === 'ASC') {
+      return (cellA.includes('$')) === false
```
ivanchuk-yurii, line 40: `cellA.includes('$')` is already boolean, you don't need to compare it with `false`

src/scripts/main.js
```diff
@@ -1,3 +1,301 @@
...
+        ? cellA.localeCompare(cellB)
+        : normalizeNumber(cellA) - normalizeNumber(cellB);
+    } else {
+      return (cellA.includes('$')) === false
+        ? cellB.localeCompare(cellA)
+        : normalizeNumber(cellB) - normalizeNumber(cellA);
```
ivanchuk-yurii, line 46: try to not duplicate it

ivanchuk-yurii (changes requested):
'office' in table should be select
https://user-images.githubusercontent.com/53611910/203856027-1187d700-ef48-40d1-aa41-e283e4e0e594.mov

tetiana-ivanchenko (changes requested):
image
sorting by salary works wrong
sorting by name, position and office doesn't work

src/scripts/main.js
AlisaVlasova: let`s try to figure it out in fe_chat, it can be faster)

AlisaVlasova (changes requested):

AlisaVlasova (approved):

PR #249, opened 2022-11-21
solution

olenitut:
- [DEMO LINK](https://<your_account>.github.io/js_employees_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -52,11 +52,7 @@ const saveChanges = function(editInput, clicked, pereviousValue) {
 
   clicked.innerHTML = '';
 
-  if (editedValue === '') {
-    clicked.textContent = pereviousValue;
-  } else {
-    clicked.textContent = editedValue;
-  }
+  clicked.textContent = editedValue || pereviousValue;
 };
 
 /// /// /// /// /// /// /// ///
@@ -81,10 +77,11 @@ const sortTable = function(e) {
         case 'Name':
         case 'Office':
           return firstText.localeCompare(secondText);
-        case 'Age':
-          return firstText - secondText;
         case 'Salary':
           return getSalaryNumber(firstText) - getSalaryNumber(secondText);
+        case 'Age':
+        default:
+          return firstText - secondText;
       };
     } else {
       switch (type) {
@@ -92,10 +89,11 @@ const sortTable = function(e) {
         case 'Name':
         case 'Office':
           return secondText.localeCompare(firstText);
-        case 'Age':
-          return secondText - firstText;
         case 'Salary':
           return getSalaryNumber(secondText) - getSalaryNumber(firstText);
+        case 'Age':
+        default:
+          return secondText - firstText;
       }
     }
   });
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,282 @@
...
+    if (sorted[index] === 0) {
+      switch (type) {
+        case 'Position':
+        case 'Name':
+        case 'Office':
+          return firstText.localeCompare(secondText);
+        case 'Age':
+          return firstText - secondText;
+        case 'Salary':
+          return getSalaryNumber(firstText) - getSalaryNumber(secondText);
```
Viktor-Kost, line 87: Better add default value to return, if cases are not met a condition

src/scripts/main.js
```diff
@@ -1,3 +1,282 @@
...
+      };
+    } else {
+      switch (type) {
+        case 'Position':
+        case 'Name':
+        case 'Office':
+          return secondText.localeCompare(firstText);
+        case 'Age':
+          return secondText - firstText;
+        case 'Salary':
```
Viktor-Kost, line 97: Better add default value to return, if cases are not met a condition

src/scripts/main.js
```diff
@@ -1,3 +1,282 @@
...
+  descriptionElement.textContent = description;
+  titleElement.classList.add('title');
+  messageElement.append(titleElement, descriptionElement);
+
+  messageElement.classList.add(type, 'notification');
+  messageElement.setAttribute('data-qa', 'notification');
+
+  document.body.append(messageElement);
+
+  setTimeout(() => {
```
Viktor-Kost, line 45: It's a good practice to add clearTimeout to avoid memory leak [mdn](https://developer.mozilla.org/en-US/docs/Web/API/clearTimeout)

src/scripts/main.js
```diff
@@ -1,3 +1,282 @@
...
+    messageElement.remove();
+  }, 3000);
+};
+
+const saveChanges = function(editInput, clicked, pereviousValue) {
+  const editedValue = editInput.value;
+
+  clicked.innerHTML = '';
+
+  if (editedValue === '') {
```
Viktor-Kost, line 55: Here we can use ternary operator

src/scripts/main.js
```diff
@@ -1,3 +1,282 @@
...
+const formElement = document.createElement('form');
+
+formElement.classList.add('new-employee-form');
+
+const inputName = createInput('Name:', 'name');
+const inputPosition = createInput('Position:', 'position');
+
+// 2) Select Office
+
+const lableSelect = document.createElement('label');
```
Viktor-Kost, line 151: Here we have a lot of similar action. 

We can create an array of cities and then using a loop (or even better an array method), we can create all of elements, setting attributes etc

`const cities = ['Tokyo', 'Singapore' .... ] `

Viktor-Kost (changes requested):

src/scripts/main.js
```diff
@@ -1,3 +1,269 @@
...
+
+  option.setAttribute('value', office);
+  option.textContent = office;
+  selecetElement.append(option);
+});
+
+lableSelect.append(selecetElement);
+formElement.append(lableSelect);
+
+// 3) Age, salary inputs
```
liliia-livandovskaya, line 170:
```suggestion
```
it's better to remove such comments, they spam your code

liliia-livandovskaya (approved):
Great job! 🔥

PR #255, opened 2022-12-05
done1

panianchuk:
https://panianchuk.github.io/js_employees_table_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,217 @@
 'use strict';
 
-// write code here
+const table = document.querySelector('table');
+const headers = table.querySelector('thead');
+const tbody = table.querySelector('tbody');
+let colIndex = -1;
+
+headers.addEventListener('click', (eventOnClick) => {
+  const target = eventOnClick.target;
+  const index = target.cellIndex;
+
+  sortingTable(index, target.innerHTML, colIndex === index);
+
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,217 @@
...
+}
+
+function toNormalNumber(string) {
+  let res = '';
+
+  string.includes('$')
+    ? res = string.slice(1).split(',').join('')
+    : res = string;
+
+  return +res;
```
loralevitska, line 50:
```suggestion
  return Number(res);
```

src/scripts/main.js
```diff
@@ -1,3 +1,217 @@
...
+    return false;
+  }
+
+  if (key === 'name' && value.length < 4) {
+    pushNotification('Error', 'error', 'Name must contain at least 4 letters');
+
+    return false;
+  }
+
+  if (key === 'age' && (+value < 18 || +value > 90)) {
```
loralevitska, line 158: It's better to set the min and max attributes for the input, this way you will avoid unnecessary checks and it will be clearer to the user

loralevitska (changes requested):
Looks great) Let's improve

1. It would be good to be in the plus on salary. Limit negative values
image

stanislav-topikha (approved):
GJ 👍🏻

PR #260, opened 2022-12-11
Develop

artem-hirzhev:
[DEMO LINK](https://artem-hirzhev.github.io/js_employees_table_DOM/)

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,4 +1,3 @@
 module.exports = {
-  extends: ['@mate-academy/eslint-config', 'plugin:cypress/recommended', 'plugin:chai-friendly/recommended'],
-  plugins: ['cypress', 'chai-friendly'],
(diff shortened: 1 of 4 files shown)


DarkMistyRoom (changes requested):
Good job! 

- will be great to add verification on editing
- what about this requirement?
image

- implement the saving on blur(editing logic)

Bohdan-Maksymiuk (changes requested):
Good job, almost done
Screenshot 2022-12-14 at 12 48 35

* would be great to have a select to edit office
* also when I click to edit, previous value should be in input, not empty input

https://user-images.githubusercontent.com/94289038/207575725-603e6c54-b473-4138-ba1e-12641b1059fa.mov

DarkMistyRoom (approved):
Great. Would be perfect to add autofocus to input on editing. like here
https://www.loom.com/share/20785a7cb52849a4b1b52f7692915946

PR #262, opened 2022-12-12
first try

RudBeee:
https://RudBeee.github.io/js_employees_table_DOM


vlad-demchuk (changes requested):
I don't know what's wrong with me(
Can you explain the type of error to a user? 
image

vlad-demchuk (changes requested):
Add an error message

PR #265, opened 2022-12-13
Solution

ifuride:
[DEMO LINK](https://ifuride.github.io/js_employees_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,264 @@
 'use strict';
 
-// write code here
+const table = document.querySelector('table');
+const tableHead = document.querySelector('thead');
+const tableBody = document.querySelector('tbody');
+const headers = document.querySelectorAll('thead > tr > th');
+const isColumnSorted = {};
+
+headers.forEach(a => {
+  isColumnSorted[a.innerText] = 1;
+});
+
+function cleanState(state) {
+  for (const headerState in state) {
+    state[headerState] = 1;
+  }
+};
+
+tableHead.addEventListener('click', (handler) => {
+  const rows = [...tableBody.rows];
+  const header = handler.target.closest('th').innerText;
+  const orderSign = isColumnSorted[header];
+
+  cleanState(isColumnSorted);
+  isColumnSorted[header] = -orderSign;
+
+  switch (header) {
+    case 'Name':
+      rows.sort((a, b) =>
+        orderSign * (a.children[0].innerText)
+          .localeCompare(b.children[0].innerText));
+      break;
+
+    case 'Position':
+      rows.sort((a, b) =>
+        orderSign * (a.children[1].innerText)
+          .localeCompare(b.children[1].innerText));
+      break;
+
+    case 'Office':
+      rows.sort((a, b) =>
+        orderSign * (a.children[2].innerText)
+          .localeCompare(b.children[2].innerText));
+      break;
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,264 @@
...
+
+    if (errorMessages.length) {
+      cell.innerText = initialValue;
+
+      pushNotification(
+        10, 10, 'There was an error:', errorMessages.join('\n'), 'error'
+      );
+    } else {
+      if (cellIndex === 4) {
+        cell.innerText = `$` + Number(cellValues[4]).toLocaleString();
```
TarasHoliuk, line 254: it should be saved in the same format:
image
there should be `,` instead of space ` `

TarasHoliuk: Then after editing and blurring this cell we'll get the error:
image

Also, it breaks down sorting

src/scripts/main.js
```diff
@@ -1,3 +1,264 @@
...
+form.addEventListener('submit', (handler) => {
+  handler.preventDefault();
+
+  const data = new FormData(form);
+  const salary = data.get('salary');
+  const employeeName = data.get('name');
+  const position = data.get('position');
+  const office = data.get('office');
+  const age = data.get('age');
+  const salaryString = `$` + Number(salary).toLocaleString();
```
TarasHoliuk, line 172: the same as below: it should be saved in the same format

src/scripts/main.js
```diff
@@ -1,3 +1,264 @@
...
+
+  input.classList.add('cell-input');
+  input.type = 'text';
+  input.name = 'cell-input';
+  input.setAttribute('value', initialValue);
+  cell.innerText = '';
+  cell.append(input);
+  input.focus();
+
+  input.addEventListener('keyup', (f) => {
```
TarasHoliuk, line 215: I know that lint doesn't allow you repeat variable names in one file, but try to rename it (`f`) somehow.

src/scripts/main.js
```diff
@@ -1,3 +1,264 @@
...
+  if (employeeName.length > 0 && employeeName.length < 4) {
+    errorMessages.push('Name should consist of at least 4 letters.');
+  }
+
+  if (age !== '' && (+age < 18 || +age > 90)) {
+    errorMessages.push('Age range should be from 18 to 90.');
+  }
+
+  if (
+    employeeName === ''
```
TarasHoliuk, line 141: You should `trim` values from string inputs, I can add employee with name that contains only spaces `    `

TarasHoliuk (changes requested):
GJ! But a few small fixes are needed

src/scripts/main.js
ifuride: I specified 'en-US' as locales parameter for the toLocaleString() method. I couldn't see this bug since my OS has English settings.

ifuride: Also added toLocaleString('en-US') here.

src/scripts/main.js
ifuride: I added trim method for the employeeName.

src/scripts/main.js
ifuride: Changed it to (handleKeypress).

nazarmatsevych (approved):
👍

PR #269, opened 2022-12-16
Sorting by title click and adding form was implemented

Andrii-Medintsev:
- [DEMO LINK](https://Andrii-Medintsev.github.io/js_employees_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,261 @@
 'use strict';
 
-// write code here
+document.body.insertAdjacentHTML('beforeend', `
+  <form class="new-employee-form">
+    <label>Name:
+      <input name="name" type="text" data-qa="name" required>
+    </label>
+
+    <label>Position:
+      <input name="position" type="text" data-qa="position" required>
+    </label>
+
+    <label>Office:
+      <select data-qa="office" name="office" required>
+        <option>Tokyo</option>;
+        <option>Singapore</option>
+        <option>London</option>
+        <option>New York</option>
+        <option>Edinburgh</option>
+        <option>San Francisco</option>
+      </select>
+    </label>
+
+    <label>Age:
+      <input name="age" type="number" data-qa="age" required>
+    </label>
+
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,261 @@
...
+      input.replaceWith(cellContent);
+    }
+
+    input.replaceWith(input.value);
+  }
+
+  input.addEventListener('keypress', (ev) => {
+    input.removeEventListener('blur', blurHandler);
+
+    if (ev.key === 'Enter' && input.value === '') {
```
loralevitska, line 253:
```suggestion
    if (ev.key === 'Enter' && !input.value) {
```

loralevitska (changes requested):
Almost done)
Fix it:
- i can add negative numbers and whitespaces by editing fields 
image
- also I can dot it by adding new employee 
image

src/scripts/main.js
```diff
@@ -1,3 +1,315 @@
...
+  if (!inputValue || inputValue.trim().length === 0) {
+    pushNotification(
+      capitalizer(inputName), ' Cannot be empty', 'error'
+    );
+
+    return false;
+  }
+
+  if (inputName === 'name' && inputValue.trim().length < 4
+  ) {
```
loralevitska, line 292: move expressions to variables

```suggestion
  if (inputName === 'name' && inputValue.trim().length < 4) {
```

src/scripts/main.js
```diff
@@ -1,3 +1,315 @@
...
+    pushNotification(
+      capitalizer(inputName), 'Should have minimum 4 letters', 'error'
+    );
+
+    return false;
+  }
+
+  if (inputName === 'age' && (inputValue < 18 || inputValue > 90)) {
```
loralevitska, line 300: same here
if you have long expression or not clear what happens here - better to move expressions to meaningful variables

loralevitska (approved):

PR #270, opened 2022-12-18
solution for js_employees_table_DOM

petrovskyi-yevhen:
- [DEMO LINK](https://petrovskyi-yevhen.github.io/js_employees_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,342 @@
 'use strict';
 
-// write code here
+const table = document.querySelector('table');
+const headers = table.querySelectorAll('th');
+const tableBody = table.querySelector('tbody');
+
+headers.forEach(function(header) {
+  header.dataset.sorted = 'DESC';
+});
+
+function sortColumn(index, isASC) {
+  let sortedRows = [];
+
+  if (isASC) {
+    if (index === 4) {
+      sortedRows = Array.from(tableBody.rows).sort((a, b) =>
+        (salaryToNumber(a.cells[index].innerHTML)
+        - salaryToNumber(b.cells[index].innerHTML)));
+    } else {
+      sortedRows = Array.from(tableBody.rows).sort((a, b) =>
+        (a.cells[index].innerHTML.localeCompare(b.cells[index].innerHTML)));
+    }
+  } else {
+    if (index === 4) {
+      sortedRows = Array.from(tableBody.rows).sort((a, b) =>
+        (salaryToNumber(b.cells[index].innerHTML)
+        - salaryToNumber(a.cells[index].innerHTML)));
+    } else {
+      sortedRows = Array.from(tableBody.rows).sort((a, b) =>
+        (b.cells[index].innerHTML.localeCompare(a.cells[index].innerHTML)));
+    }
+  }
+
+  table.tBodies[0].append(...sortedRows);
+}
+
+headers.forEach(function(header, index) {
+  header.addEventListener('click', (e) => {
+    let order;
+
+    if (e.target.dataset.sorted === 'ASC') {
+      e.target.dataset.sorted = 'DESC';
+      order = false;
+    } else {
+      e.target.dataset.sorted = 'ASC';
+      order = true;
+    }
+    sortColumn(index, order);
+  });
+});
+
+const pushNotification = (posTop, posRight, title, description, type) => {
+  document.body.insertAdjacentHTML('afterbegin', `
+    <div data-qa="notification" class="notification ${type}"
+      style="top: ${posTop}px; right: ${posRight}px;">
+      <h2 class="title">${title}</h2>
+      <p>${description}</p>
+    </div>
+    `);
+
+  const notification = document.querySelector('.notification');
+
+  setTimeout(() => notification.remove(), 2000);
+};
+
+table.insertAdjacentHTML('afterend', `
+  <form class="new-employee-form">
+    <label>Name: 
+      <input
+        data-qa="name"
+        name="name"
+        type="text"
+        required
+      >
+    </label>
+    <label>Position: 
+      <input
+        data-qa="position"
+        name="position"
+        type="text"
+        required
+      >
+    </label>
+    <label>Office: 
+      <select data-qa="office" name="office" type="text">
+        <option value="Tokyo">Tokyo</option>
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,342 @@
 'use strict';
 
-// write code here
+const table = document.querySelector('table');
+const headers = table.querySelectorAll('th');
+const tableBody = table.querySelector('tbody');
+
+headers.forEach(function(header) {
```
polosanya, line 7:
```suggestion
headers.forEach((header) => {
```
Don't use function declaration in callbacks. Better use arrow functions instead
https://mate-academy.github.io/style-guides/javascript.html#81-when-you-must-use-an-anonymous-function-as-when-passing-an-inline-callback-use-arrow-function-notation

src/scripts/main.js
```diff
@@ -1,3 +1,342 @@
...
+    if (index === 4) {
+      sortedRows = Array.from(tableBody.rows).sort((a, b) =>
+        (salaryToNumber(a.cells[index].innerHTML)
+        - salaryToNumber(b.cells[index].innerHTML)));
+    } else {
+      sortedRows = Array.from(tableBody.rows).sort((a, b) =>
+        (a.cells[index].innerHTML.localeCompare(b.cells[index].innerHTML)));
+    }
+  } else {
+    if (index === 4) {
```
polosanya, line 24: Try to avoid nesting if...else. Please rewrite this part without it. You can use a ternary operator for that

src/scripts/main.js
```diff
@@ -1,3 +1,342 @@
...
+}
+
+headers.forEach(function(header, index) {
+  header.addEventListener('click', (e) => {
+    let order;
+
+    if (e.target.dataset.sorted === 'ASC') {
+      e.target.dataset.sorted = 'DESC';
+      order = false;
+    } else {
```
polosanya, line 44: Same here. You can easily use ternary operator here

src/scripts/main.js
```diff
@@ -1,3 +1,342 @@
...
+      e.target.dataset.sorted = 'ASC';
+      order = true;
+    }
+    sortColumn(index, order);
+  });
+});
+
+const pushNotification = (posTop, posRight, title, description, type) => {
```
polosanya, line 52: Why do you use an arrow function here and a function declaration above? Choose something one in order not to confuse other developers

polosanya (changes requested):
Good job! Let's improve your code a bit)

src/scripts/main.js
```diff
@@ -1,3 +1,340 @@
...
+
+headers.forEach((header) => {
+  header.dataset.sorted = 'DESC';
+});
+
+function sortColumn(index, isASC) {
+  let sortedRows = [];
+
+  if (isASC) {
+    // eslint-disable-next-line chai-friendly/no-unused-expressions
```
Anastasiia-Svintsova, line 15: Don't disable eslint

src/scripts/main.js
```diff
@@ -1,3 +1,340 @@
...
+      : sortedRows = Array.from(tableBody.rows).sort((a, b) =>
+        (a.cells[index].innerHTML.localeCompare(b.cells[index].innerHTML)));
+  } else {
+    // eslint-disable-next-line chai-friendly/no-unused-expressions
+    index === 4
+      ? sortedRows = Array.from(tableBody.rows).sort((a, b) =>
+        (salaryToNumber(b.cells[index].innerHTML)
+        - salaryToNumber(a.cells[index].innerHTML)))
+      : sortedRows = Array.from(tableBody.rows).sort((a, b) =>
+        (b.cells[index].innerHTML.localeCompare(a.cells[index].innerHTML)));
```
Anastasiia-Svintsova, line 29: You use the same logic here. Maybe you need just to create additional variables for a and b and switch them if is not `ASC`. For example
```
Array.from(tableBody.rows).sort((a, b) => {
  const first = isASC ? a : b;
  const second = isASC ? b : a;

  return index === 4 ? ... : ...
}

It's just an example. Maybe you'll find a better solution

File: src/scripts/main.js [hunk 7]
```diff
@@ -1,3 +1,340 @@
...
+  }
+
+  table.tBodies[0].append(...sortedRows);
+}
+
+headers.forEach((header, index) => {
+  header.addEventListener('click', (e) => {
+    let order;
+
+    // eslint-disable-next-line chai-friendly/no-unused-expressions
```
Anastasiia-Svintsova, line 39: Don't disable eslint. Try to solve the error

src/scripts/main.js
```diff
@@ -1,3 +1,340 @@
...
+    e.target.dataset.sorted === 'ASC'
+      ? (e.target.dataset.sorted = 'DESC',
+      order = false)
+      : (e.target.dataset.sorted = 'ASC',
+      order = true);
```
Anastasiia-Svintsova, line 44: fix this 
image

src/scripts/main.js
```diff
@@ -1,3 +1,340 @@
...
+  input.addEventListener('keydown', (evt) => {
+    if (evt.code !== 'Enter') {
+      return;
+    }
+
+    inputValidation(index);
+  });
+
+  const inputValidation = (i) => {
+    if (input.value !== '') {
```
Anastasiia-Svintsova, line 300: To avoid nesting you can return what you need if you don't have `input.value` in the begining
```
if (!input.value) {
  return row.innerText = tempValue;
}

Anastasiia-Svintsova (changes requested):
- I shouldn't be able to add an employee with empty fields

https://user-images.githubusercontent.com/99403198/209542644-3635e327-f9ba-4a16-a839-1d97c58f1a40.mov

- Messages look a little bit strange. Would be better to rewrite it in English
image

- The office shouldn't change to Tokyo by default while editing

https://user-images.githubusercontent.com/99403198/209542838-a0900223-9ce9-4a79-aa69-9b60fc8a8f0f.mov

src/scripts/main.js
petrovskyi-yevhen: Якщо робити так, то лінтер матюкається

qoa11a (approved):

PR #273, opened 2022-12-26
My solution

Vlad-Fedorishchev:
Demo Link  -  https://Vlad-Fedorishchev.github.io/js_employees_table_DOM/


polosanya (changes requested):
1. Would be great to specify why exactly the name is not valid. Same with other errors
image
Also, it seems like errors should disappear after some time, how it was in previous tasks
2. There should be an ability to select an office on edit
image

SanyaBratashchuk (changes requested):
Add validation for edit 
image

Vlad-Fedorishchev:
@qoa11aI did data entry checks when editing table cells, why do you write that I did not validate? But if the data is entered incorrectly, you will see an error.

marina-tilniak (approved):

PR #274, opened 2022-12-29
solution

Deniskakaka:


TarasHoliuk:
Formatting on add / edit items should be the same as you do on sort:

after adding:
image

and then after sorting (also it shouldn't add `,000` in the end, just format)
image

TarasHoliuk:
- clear form after adding new item

src/scripts/main.js
```diff
@@ -1,3 +1,315 @@
 'use strict';
 
-// write code here
+const body = document.querySelector('tbody');
```
TarasHoliuk, line 3:
```suggestion
const tableBody = document.querySelector('tbody');
```

src/scripts/main.js
```diff
@@ -1,3 +1,315 @@
...
+      }
+
+      if (numbers.includes(sortKey)) {
+        return secondElement - firstElement;
+      }
+    };
+  };
+};
+
+sortBy.forEach(el => {
```
TarasHoliuk, line 185: use [eventDelagation](https://javascript.info/event-delegation)

TarasHoliuk (changes requested):

src/scripts/main.js
OlEzhA0: not fixed
why do we need to add listener to each element when you can use event delegation

OlEzhA0:
Screenshot 2023-01-03 at 4 33 29 PM
I shouldn’t have possibility to add negative age

OlEzhA0 (changes requested):
hey! Good job!
Please check the article on how delegation works and apply it to your project

mykhalenych (approved):

PR #275, opened 2022-12-29
Employee table

sergey-vorobei:
- [DEMO LINK](https://sergey-vorobei.github.io/js_employees_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,228 @@
 'use strict';
 
-// write code here
+const tHead = document.querySelector('thead');
+const tBody = document.querySelector('tbody');
+
+let asc = true;
+let checkColumn;
+
+function ascSort(rowCollection, targetIndex) {
+  asc = true;
+
+  return Array.from(rowCollection).sort((a, b) => {
+    const cellA = a.cells[targetIndex].innerText;
+    const cellB = b.cells[targetIndex].innerText;
+
+    return isNaN(normalize(cellA))
+      ? cellA.localeCompare(cellB)
+      : normalize(cellA) - normalize(cellB);
+  });
+}
+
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,228 @@
...
+      input.remove();
+    }
+  };
+
+  cell.addEventListener('keydown', (evt) => {
+    if (evt.key === 'Enter') {
+      if (input.value.length <= 0) {
+        cell.textContent = cellPrevValue;
+      } else {
+        cell.textContent = input.value;
```
polosanya, line 224: You can use a ternary operator here to make your code cleaner

src/scripts/main.js
```diff
@@ -1,3 +1,228 @@
...
+    return pushNotification('Error',
+      'Please, enter at least 4 characters', 'error');
+  } else if (positionInput.length <= 0) {
+    return pushNotification('Error',
+      'Please, enter position', 'error');
+  } else if (ageInput < 18 || ageInput > 90) {
+    return pushNotification(
+      'Error', 'Please, enter a valid age', 'error');
+  } else {
+    const row = document.createElement('tr');
```
polosanya, line 158: There is no need for `else...if`. The code will not go further after `return` anyway.
Also, the code looks a bit complicated with `else...if`

polosanya (changes requested):
1. I think it would be nice to clarify which input is the problem
image
2. Is it a bug or a feature?))
error

SanyaBratashchuk (changes requested):
unify edit and create validation rules. E.g. I can not add employee with name shorter than 4 letters but can edit it like this

qoa11a (approved):

PR #279, opened 2023-01-04
Solution

offsset:
https://offsset.github.io/js_employees_table_DOM/


vlad-demchuk (changes requested):
fix validation, I can add wrong data now, and can't add if my name has 4 letters
https://user-images.githubusercontent.com/102411130/210748488-fa36e8c2-e4f7-4c64-8a0f-596c34762b0d.mov

vlad-demchuk (changes requested):
now i can't set age
image

nazarmatsevych (approved):

PR #280, opened 2023-01-04
solution

Ishkina:
[DEMO LINK](https://ishkina.github.io/js_employees_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,270 @@
 'use strict';
 
-// write code here
+const body = document.querySelector('body');
+const tabHead = document.querySelector('thead');
+const tabBody = document.querySelector('tbody');
+
+const rows = tabBody.rows;
+
+let clickCount;
+
+function convertToNumber(elem) {
+  return elem.split('$').join('').split(',').join('');
+}
+
+function pushNotification(title, description, type) {
+  const div = document.createElement('div');
+
+  div.classList.add('notification', type);
+  div.setAttribute('data-qa', 'notification');
+
+  div.insertAdjacentHTML('afterbegin', `
+    <h1 class="title">${title}</h1>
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,270 @@
 'use strict';
 
-// write code here
+const body = document.querySelector('body');
+const tabHead = document.querySelector('thead');
+const tabBody = document.querySelector('tbody');
+
+const rows = tabBody.rows;
+
+let clickCount;
```
vlad-demchuk, line 9: let's assign some value at once. just assign `null` if you don't have some needed value yet

src/scripts/main.js
```diff
@@ -1,3 +1,270 @@
...
+      if (isNaN(first)) {
+        return first.localeCompare(second);
+      } else {
+        return first - second;
+      }
+    } else if (clickCount === 2) {
+      if (isNaN(second)) {
+        return second.localeCompare(first);
+      } else {
+        return second - first;
```
vlad-demchuk, line 54: it will be clear to use switch case here

vlad-demchuk (changes requested):
don't reset form when data is wrong

Uploading Screen Recording 2023-01-05 at 10.44.28.mov…

nazarmatsevych (changes requested):
Try to fix all of the previos comment before making re-request.
If you have any questions - feel free to ask them in fe_chat

roman-mirzoian (changes requested):
Good progress!
But when I try to sort the table, an error appears in the console:
Image

Ishkina:
I may have added the update incorrectly, since the sorting is working for me and no error appears in the console. I'm not sure what needs to be fixed.

brespect (approved):
Well done!

PR #286, opened 2023-01-10
Solution

Vasyliy22:
[DEMO LINK](https:Vasyliy22.github.io/js_employees_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,226 @@
 'use strict';
 
-// write code here
+const body = document.querySelector('body');
+const table = body.querySelector('table');
+const tableBody = table.querySelector('tbody');
+const headers = table.querySelectorAll('th');
+const rows = tableBody.querySelectorAll('tr');
+const form = document.createElement('form');
+let sortCol;
+let sortAsc = false;
+
+[].forEach.call(headers, (header, index) => {
+  header.addEventListener('click', (events) => {
+    sortColumn(index, false);
+
+    // const curr = rows;
+
+    // rows.forEach(item => {
+    //   if (item === curr) {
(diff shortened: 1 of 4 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,226 @@
...
+        return sortAsc
+          ? convertToNumber(cellA) - convertToNumber(cellB)
+          : convertToNumber(cellB) - convertToNumber(cellA);
+      }
+
+      return sortAsc
+        ? cellA.localeCompare(cellB)
+        : cellB.localeCompare(cellA);
+    } else {
+      return sortAsc
```
polosanya, line 80: Do you really need else here? Your code will not go further after return anyway

polosanya (changes requested):
Please fix the link so we can check your work properly

Vasyliy22:
[DEMO LINK](https://Vasyliy22.github.io/js_employees_table_DOM/)

Anastasiia-Svintsova (changes requested):
- Clarify the error message, please
image
- I can write not valid values to the age field, fix it, please
image
- Clear the form after submission. And new employees should take part in a sorting function

https://user-images.githubusercontent.com/99403198/212302693-f7ea512f-5a07-4234-bbdd-0f64f63db8c1.mov

- Fix the editing according to the task, please

https://user-images.githubusercontent.com/99403198/212302973-92370cad-8e8a-44b6-acc2-79bc0c810f2b.mov

qoa11a (changes requested):
image
Fix this moment please. If you return something inside if the code won't go further so there's no need to keep this else.

qoa11a (approved):

PR #287, opened 2023-01-11
solution without edit cell

webdevw:
- [DEMO LINK](https://webdevw.github.io/js_employees_table_DOM/)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 1. Replace `<your_account>` with your Github username in the link
-    - [DEMO LINK](https://<your_account>.github.io/js_employees_table_DOM/)
+    - [DEMO LINK](https://webdevw.github.io/js_employees_table_DOM/)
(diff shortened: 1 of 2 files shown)


polosanya (changes requested):
Almost done! Let's make the error text more descriptive
1. Would be great to specify in the error text which format is required
image
Also, error text doesn't disappear
2. Success message also doesn't disappear.
image

Anastasiia-Svintsova (changes requested):
I can add an employee with empty Name and Position fields. Fix it, please

https://user-images.githubusercontent.com/99403198/212314782-bc84d474-13f1-40e4-a113-05f56e720924.mov
- Seems that you forgot to implement this part of the task 
image

https://user-images.githubusercontent.com/99403198/212315165-7c1ec6c1-a930-47ed-9408-08ae8f8c15f6.mov

polosanya (approved):

PR #288, opened 2023-01-11
solution

AnnyKoval:
- [DEMO LINK](https://AnnyKoval.github.io/js_employees_table_DOM/)

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,4 +1,3 @@
 module.exports = {
-  extends: ['@mate-academy/eslint-config', 'plugin:cypress/recommended', 'plugin:chai-friendly/recommended'],
-  plugins: ['cypress', 'chai-friendly'],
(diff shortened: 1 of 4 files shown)


Anastasiia-Svintsova (changes requested):
- Add a min value for this input 
image
- Now I can add an employee with empty fields, fix it, please

https://user-images.githubusercontent.com/99403198/212051236-6d7aeeef-3242-41b5-8a3c-1d9afa1742a5.mov

- Check if values are valid before saving it after editing
image

AnnyKoval:
how can I do this without "required" in HTML? If i set up || fields[0].children[0].value === null - nothing happened

mykhalenych (approved):

mykhalenych:
how can I do this without "required" in HTML? If i set up || fields[0].children[0].value === null - nothing happened

document.getElementById("test").required = true;

PR #292, opened 2023-01-15
Solution

tarasavage:
[DEMO LINK](https://tarasavage.github.io/js_employees_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -166,7 +166,7 @@ body.addEventListener('dblclick', (evt) => {
 
       case 4:
         newCell.value = newCell.value !== ''
-          ? '$' + Number(newCell.value).toLocaleString('en-US')
+          ? '$' + Number(newCell.value.remove('$')).toLocaleString('en-US')
           : previousValue;
 
         break;
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,211 @@
 'use strict';
 
-// write code here
+const header = document.querySelector('thead');
+const body = document.querySelector('tbody');
+
+let counter = 0;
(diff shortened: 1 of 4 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,211 @@
...
+});
+
+body.addEventListener('dblclick', (evt) => {
+  const selectedCell = evt.target;
+  const previousValue = selectedCell.textContent;
+  const newCell = document.createElement('input');
+
+  newCell.classList.add('cell-input');
+
+  switch (selectedCell.cellIndex) {
```
witflash, line 132: You could use an `if / else` block instead of `switch / case`. You don't need to use it for two cases.

src/scripts/main.js
```diff
@@ -1,3 +1,211 @@
...
+          return numX - numY;
+
+        default:
+          return 0;
+      }
+    });
+
+  counter = indexTitle === currentIndex
+    ? counter
+    : 0;
```
witflash, line 39:
```
if (indexTitle !== currentIndex) {
  counter = 0;
}
```

src/scripts/main.js
```diff
@@ -1,3 +1,211 @@
...
+
+        break;
+
+      default:
+        if (newCell.value === '') {
+          newCell.value = previousValue;
+        }
+
+        break;
+    }
```
witflash, line 180: It's a very complex block for me. 
Could you use only one approach (`switch/case` or `if/else`)?

witflash (changes requested):
Good job!
Check my comments, please.

Viktor-Kost (approved):
Nice!

PR #296, opened 2023-01-26
Develop

U-Ra-All:
- [DEMO LINK](https://U-Ra-All.github.io/js_employees_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -71,11 +71,11 @@ function selectRow(table) {
 }
 
 function createForm(table) {
-  const form = document.createElement('form');
+  const newEmployeeForm = document.createElement('form');
 
-  form.classList.add('new-employee-form');
+  newEmployeeForm.classList.add('new-employee-form');
 
-  form.innerHTML = `
+  newEmployeeForm.innerHTML = `
     <label>
       Name: 
       <input 
@@ -103,7 +103,7 @@ function createForm(table) {
         data-qa="office"
         required
       >
-        <option>Tokyo</option>
+        <option value="default">Tokyo</option>
         <option>San Francisco</option>
         <option>Singapore</option>
         <option>New York</option>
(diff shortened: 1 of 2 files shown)


vlad-demchuk (changes requested):
implement this point
image

all fields should be required (I can add new employee without age and salary now)
image
image

polosanya (changes requested):
The previous comment is not fixed. I still can create an employee without age and salary.
image
And why did Tokyo change to 'default'? Just leave it Tokyo if user chooses nothing

U-Ra-All:
> The previous comment is not fixed. I still can create an employee without age and salary. image And why did Tokyo change to 'default'? Just leave it Tokyo if user chooses nothing

Thanks, done

src/scripts/main.js
```diff
@@ -1,3 +1,302 @@
...
+  });
+
+  input.addEventListener("blur", (e) => {
+    setEditedCellValue();
+  });
+
+  function setEditedCellValue() {
+    if (isEditing) {
+      editedCell.textContent =
+        input.value.trim() !== "" ? input.value.trim() : previousTextContent;
```
TarasHoliuk, line 298:
```suggestion
        input.value.trim() || previousTextContent;
```

TarasHoliuk: also, it would be nice to format the value (salary) after editing

image

TarasHoliuk (changes requested):
Nicely done :+1: But let's improve it a bit

src/scripts/main.js
U-Ra-All: Done

Viktor-Kost (approved):
Ok, it is good) 
I have added one more recommendation just for you to use

src/scripts/main.js
```diff
@@ -1,3 +1,319 @@
...
+              ).toLocaleString('en-US')}`
+              : `$${parseInt(
+                input.value.trim().split(',').join('')
+              ).toLocaleString('en-US')}`;
+        } else {
+          editedCell.textContent = input.value.trim();
+        }
+      } else {
+        editedCell.textContent = previousTextContent;
+      }
```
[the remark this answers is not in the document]
U-Ra-All: Ok, thanks!

PR #297, opened 2023-01-26
Commit

skomskiy:
- [DEMO LINK](https://skomskiy.github.io/js_employees_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,285 @@
 'use strict';
 
-// write code here
+const thead = document.querySelector('thead');
+const tbody = document.querySelector('tbody');
+const tr = tbody.rows;
+const th = thead.querySelectorAll('th');
+
+thead.addEventListener('click', e => {
+  if (e.target.getAttribute('data-sorted') === 'ASC') {
+    if (e.target.textContent === 'Name') {
+      const sortNames = function(arr) {
+        return [...arr].sort((a, b) =>
+          b.children[0].textContent < a.children[0].textContent ? -1 : 1);
+      };
+
+      tbody.append(...sortNames(tr));
+    };
+
+    if (e.target.textContent === 'Position') {
+      const sortNames = function(arr) {
+        return [...arr].sort((a, b) =>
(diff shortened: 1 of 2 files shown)


SanyaBratashchuk (changes requested):
How can I edit employee?

src/scripts/main.js
```diff
@@ -1,3 +1,403 @@
...
+    }
+
+    e.target.setAttribute('data-sorted', 'ASC');
+  }
+});
+
+tbody.addEventListener('click', e => {
+  const patentTarget = e.target.parentElement;
+  // const active = document.getElementsByClassName('active');
+  // console.log(active);
```
Volodymyr-Melnyk-Fron10der, line 119: Remove all comments

src/scripts/main.js
```diff
@@ -1,3 +1,403 @@
...
+
+    return;
+  }
+
+  if (ageInput.value < 18 || ageInput.value > 90) {
+    pushNotification(
+      'Erorr',
+      'Age should be between 18-90',
+      'error'
+    );
```
Volodymyr-Melnyk-Fron10der, line 247: Don't repeat yourself, move all magic string into variables 'Error' 'error' or create function 'pushError' which takes only error message

Volodymyr-Melnyk-Fron10der (changes requested):
Great, just few fixes

skomskiy:
- [DEMO LINK](https://skomskiy.github.io/js_employees_table_DOM/)

ViktorHavryliuk (changes requested):
Please use switch statement for sorting logic instead of multiple ifs
Also, using order check as root if statement brings too much repeating code, as you basically do same things in "else" but with opposite operation. Try to find a better solution and feel free to ask in chat if you need help
```
switch (e.target.textContent){
  case 'Name':
    <sort names>
    break;
  ...
}
```

skomskiy:
- [DEMO LINK](https://skomskiy.github.io/js_employees_table_DOM/)

ivbutsykin (approved):

PR #298, opened 2023-01-26
table & form added

anna-akhalmosulishvili:
https://<anna-akhalmosulishvili>.github.io/js_employees_table_DOM/

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 1. Replace `<your_account>` with your Github username in the link
-    - [DEMO LINK](https://<your_account>.github.io/js_employees_table_DOM/)
+    - [DEMO LINK](https://<anna-akhalmosulishvili>.github.io/js_employees_table_DOM/)
 2. Follow [this instructions](https://mate-academy.github.io/layout_task-guideline/)
     - Run `npm run test` command to test your code;
     - Run `npm run test:only -- -n` to run fast test ignoring linter;
@@ -41,11 +41,11 @@ Start table:
 ```
 - Add qa attributes for each input field:
 ```
- data-qa="name" 
- data-qa="position" 
- data-qa="office" 
- data-qa="age" 
- data-qa="salary" 
+ data-qa="name"
+ data-qa="position"
+ data-qa="office"
+ data-qa="age"
+ data-qa="salary"
 ```
 - Select should have 6 options: `Tokyo`, `Singapore`, `London`, `New York`, `Edinburgh`, `San Francisco`.
 - Use texts for labels and buttons from the screenshot below.
@@ -61,7 +61,7 @@ Start table:
 - Notification titles and descriptions are up to you.
 - Add qa attribute for notification: `data-qa="notification"` and class `error`/`success` depending on the result.
 
-##### Implement editing of table cells by double-clicking on it (optional). 
+##### Implement editing of table cells by double-clicking on it (optional).
 - Double click on the cell of the table, should remove text, and append input with `cell-input` class.
 - The input value should contain replaced by input text.
 - Only one cell can be edited at the time.
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,346 @@
 'use strict';
 
-// write code here
+const body = document.querySelector('body');
+const head = document.querySelector('thead');
+const titles = [...head.children[0].children];
+const tbody = document.querySelector('tbody');
+
+// #region --Sorting table--
+
+function sortTable(index) {
+  const table = document.querySelector('tbody');
+  let isSwitching = true;
+  let countSwitch = 0;
+  let dir = 'asc';
+
+  while (isSwitching) {
+    isSwitching = false;
+
+    const rows = table.rows;
+    let shouldRowsSwitch = false;
+    let i;
+
+    for (i = 0; i < rows.length - 1; i++) {
+      shouldRowsSwitch = false;
+
+      const firstRow = rows[i].getElementsByTagName('TD')[index];
+      const secondRow = rows[i + 1].getElementsByTagName('TD')[index];
+
+      const firstRowContent = firstRow.innerHTML;
+      const secondRowContent = secondRow.innerHTML;
+
+      const isFirstRowNum = Number(firstRowContent);
+      const isSecondRowNum = Number(secondRowContent);
+
+      if (dir === 'asc') {
+        if (isFirstRowNum) {
+          if (isFirstRowNum > isSecondRowNum) {
+            shouldRowsSwitch = true;
+            break;
+          }
+        } else {
+          const isCurrency = firstRowContent.indexOf('$');
+
+          if (isCurrency !== -1) {
+            const firstCurrency = +firstRowContent.slice(1).split(',').join('');
+            const secCurrency = +secondRowContent.slice(1).split(',').join('');
+
+            if (firstCurrency > secCurrency) {
+              shouldRowsSwitch = true;
+              break;
+            }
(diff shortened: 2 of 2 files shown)


SanyaBratashchuk:
> https://.github.io/js_employees_table_DOM/

Pay more attention to links that you provide

README.md
```diff
@@ -1,5 +1,5 @@
 1. Replace `<your_account>` with your Github username in the link
-    - [DEMO LINK](https://<your_account>.github.io/js_employees_table_DOM/)
+    - [DEMO LINK](https://<anna-akhalmosulishvili>.github.io/js_employees_table_DOM/)
```
SanyaBratashchuk, line 2:
```suggestion
    - [DEMO LINK](https://anna-akhalmosulishvili.github.io/js_employees_table_DOM/)
```

SanyaBratashchuk (changes requested):
How can I edit employee?

README.md
```diff
@@ -1,5 +1,5 @@
 1. Replace `<your_account>` with your Github username in the link
-    - [DEMO LINK](https://<your_account>.github.io/js_employees_table_DOM/)
+    - [DEMO LINK](https://anna-akhalmosulishvili.github.io/js_employees_table_DOM/)
```
DarkMistyRoom, line 2: paste this link to the description

DarkMistyRoom (changes requested):
- I can't select another row before clicking on selected one

https://www.loom.com/share/1ce7bf532a824f90a8487b3a0a7169ae

- reset form on sucsess

- what about this requrement?
image

- add validation to editing

src/scripts/main.js
```diff
@@ -1,3 +1,429 @@
 'use strict';
 
-// write code here
+const body = document.querySelector('body');
+const head = document.querySelector('thead');
+const titles = [...head.children[0].children];
+const tbody = document.querySelector('tbody');
+
+// #region --Sorting table--
```
loralevitska, line 8: Remove comments

src/scripts/main.js
```diff
@@ -1,3 +1,429 @@
...
+
+function sortTable(index) {
+  const table = document.querySelector('tbody');
+  let isSwitching = true;
+  let countSwitch = 0;
+  let dir = 'asc';
```
loralevitska, line 14: What does it mean `dir`?  Use more meaningful variable names

src/scripts/main.js
```diff
@@ -1,3 +1,429 @@
...
+
+      const firstRowContent = firstRow.innerHTML;
+      const secondRowContent = secondRow.innerHTML;
+
+      const isFirstRowNum = Number(firstRowContent);
+      const isSecondRowNum = Number(secondRowContent);
+
+      if (dir === 'asc') {
+        if (isFirstRowNum) {
+          if (isFirstRowNum > isSecondRowNum) {
```
loralevitska, line 37: Try to avoid such nesting. Use helper functions

src/scripts/main.js
```diff
@@ -1,3 +1,429 @@
...
+
+// #endregion
+
+// #region --Select a row--
+
+tbody.addEventListener('click', e => {
+  const row = e.target.parentNode;
+  const activeElements = [...document.querySelectorAll('.active')];
+
+  if (activeElements.length === 0) {
```
loralevitska, line 121:
```suggestion
  if (!activeElements.length) {
```

src/scripts/main.js
```diff
@@ -1,3 +1,429 @@
...
+      warningPopUp.style.display = 'none';
+    }, 2000);
+  }
+
+  const cellIndex = [...cell.parentNode.children].indexOf(cell);
+
+  if (((cellIndex === 0)
+  && ((newCellValue.length > 0) && (newCellValue.length < 4)))
+  || ((cellIndex === 3)
+  && (newCellValue.length > 0) && (newCellValue < 18 || newCellValue >= 90))) {
```
loralevitska, line 393: Move to variables such long expressions

loralevitska (changes requested):
Please review the comments and correct your code in all places relevant to the comment, not just in the comment place, also avoid duplicating code. Everything that is repeated should be transferred to auxiliary functions or variables. If you have questions or difficulties, ask in the chat

Keep up the good work! Some comments to fix:

1. Use `string.trim()` to avoid empty values (in form and edit in table too)
image
2. Salary should more than 1

src/scripts/main.js
```diff
@@ -1,3 +1,495 @@
...
+
+      const isFirstRowNum = Number(firstRowContent);
+      const isSecondRowNum = Number(secondRowContent);
+
+      if (sortDirection === 'asc') {
+        if (isFirstRowNum && (isFirstRowNum > isSecondRowNum)) {
+          shouldRowsSwitch = true;
+          break;
+        } else {
+          const isCurrency = firstRowContent.indexOf('$');
```
ViktorHavryliuk, line 64: isCurrency name suggest that variable is boolean type. It would be better to either make check here or rename variable
```suggestion
          const isCurrency = firstRowContent.indexOf('$') !== -1;
```

src/scripts/main.js
```diff
@@ -1,3 +1,495 @@
...
+            if (compareCurrency(firstRowContent, secondRowContent)) {
+              shouldRowsSwitch = true;
+              break;
+            };
+          } else {
+            if (compareStrings(firstRowContent, secondRowContent)) {
+              shouldRowsSwitch = true;
+              break;
+            }
+          }
```
ViktorHavryliuk, line 76: A lot of dublication here, try to find a way to get rid of one if, at least

src/scripts/main.js
```diff
@@ -1,3 +1,495 @@
...
+          } else {
+            if (compareStrings(firstRowContent, secondRowContent) === false) {
+              shouldRowsSwitch = true;
+              break;
+            }
+          }
+        }
+      }
+    }
+
```
ViktorHavryliuk, line 101: This almost exaclty dublicates if block above. Consider reorganising your algorithm, changing conditions and their order, maybe creating some helper functions or modifying existing ones and try to find a way to fix this

src/scripts/main.js
```diff
@@ -1,3 +1,495 @@
...
+    let inputValue = label.children[0].value;
+
+    inputValue = inputValue.trim();
+
+    if (checkLength(inputValue)) {
+      showWarning();
+      incorrectData = true;
+    }
+
+    switch (i) {
```
ViktorHavryliuk, line 412: It would be better to get input's name attribute and switch based on that, it's much easier to understand code like
```
switch (inputName) {
  case "name":
  ...
}
```
than when everything is index based
Please fix in other cases too

ViktorHavryliuk (changes requested):
Try to find a way to reduce code duplication and make it easier to understand. For example, one trick you could apply to make sorting in needed order shoter is to use Array.sort() method first for ASC order, and then, if you need DESC, just use Array.reverse(). Example:
```
const sortedValues = [...values].sort(sortString)

if (sortDirection === 'desc') {
  sortedValues.reverse():
}

// update your rows based on sortedValues array
```

slavanosatskyi (changes requested):
You did a lot of nice work, but still there is some unfinished tasks.
1. There are cases when sorting doesn't work, please check a video
https://user-images.githubusercontent.com/6526149/217868794-b5cbf1f9-2b7f-4958-8a2f-4384a6aa8e76.mov
2. Still it is not possible to edit users in the table

If you have any questions about this task, feel free to ask us in corresponding chats, we will be glad to help you

PR #301, opened 2023-01-27
first_try

Vadym-Prydatok:
- [DEMO LINK](https://Vadym-Prydatok.github.io/js_employees_table_DOM/)


src/scripts/main.js
```diff
@@ -1,3 +1,300 @@
...
+  <option>San Francisco</option>
+`;
+newButton.textContent = 'Save to table';
+newButton.id = 'addEmployees';
+newForm.id = 'addForm';
+
+newForm.className = 'new-employee-form';
+document.body.children[0].after(newForm);
+
+for (let i = 1; i <= 5; i++) {
```
ViktorHavryliuk, line 77: It's not the clearest way to create inputs. Would be better to create a function for it and call it for each input, like:
```
createInput("Name", { name: 'name', type: 'text', ...etc });
createInput("Position", { name: 'position', type: 'text', ...etc })
```
This is just an example of one way how that function could work, you don't have to implement it exactly like this.

src/scripts/main.js
```diff
@@ -1,3 +1,300 @@
...
+      pushNotification(150, 10, `${titleSort.children[0]
+        .children[index].textContent} is to short!`,
+      `You enter ${info.length} characters.\n `
+      + 'Please enter more than 4 characters.', 'error');
+
+      return false;
+    }
+  }
+
+  if (index === 3) {
```
ViktorHavryliuk, line 237: This way it's hard to understand which field you are checking, don’t use indexes. Better check by input name, and make separate functions for each validation:
```
switch (name) {
  case 'name':
    validateName(info);
    break;
  case 'position':
    validatePosition(info);
    break;
} 
```
Also, would be better to change 'info' to 'value', as info is too broad and may mean other things
And use switch instead of multiple ifs

ViktorHavryliuk (changes requested):

src/scripts/main.js
Vadym-Prydatok: Найзрозумілішим напевно було б просто написати html код через innerhtml, але то не цікаво, тому написав через функцію, як ви порадили

src/scripts/main.js
```diff
@@ -1,3 +1,316 @@
...
+
+      return;
+    }
+    newCell.parentElement.textContent = newCell.value;
+
+    return;
+  }
+
+  newCell.parentElement.textContent = oldInfo.textContent;
+};
```
DarkMistyRoom, line 220: would be great to refactor this part with plane ifs

DarkMistyRoom (changes requested):

PR #303, opened 2023-02-01
Solution

yana-kotsulym:


yana-kotsulym:
https://yana-kotsulym.github.io/js_employees_table_DOM/

src/scripts/main.js
```diff
@@ -1,3 +1,291 @@
...
+
+let index;
+let count = 1;
+
+tableHead.addEventListener('click', (e) => {
+  const item = e.target.textContent.toLowerCase();
+
+  const dir = count % 2 === 0 && item === index
+    ? 'desc'
+    : 'abs';
```
loralevitska, line 180: What does it mean `abs`?
```suggestion
    : 'asc';
```

loralevitska (changes requested):
- Remove this 
image
- Use `trim()` to avoid empty values and validate number values (numbers must be positive)
image
- save edited values 

https://user-images.githubusercontent.com/70009205/216373116-6cc31d53-a0c2-4efe-9c7e-49996135b786.mov

- For `office` column (for edit in row) should be select

Volodymyr-Melnyk-Fron10der (changes requested):
Hi, 

1. Office is still not saved after sorting (previous comment from Lora)
2. Smth weird happens here😅

https://user-images.githubusercontent.com/122989585/217770744-c0b294de-ccbb-422c-88e3-2c9c4b5a5596.mp4

src/scripts/main.js
```diff
@@ -1,3 +1,321 @@
...
+          <option value="Tokyo">Tokyo</option>
+          <option value="Singapore">Singapore</option>
+          <option value="London">London</option>
+          <option value="New York">New York</option>
+          <option value="Edinburgh">Edinburgh</option>
+        </select>
+      </label>
+      <label>Age:
+        <input name="age" type="number" data-qa="age">
+      </label>
```
slavanosatskyi, line 127:
```suggestion
      <label>
        Age:
        <input name="age" type="number" data-qa="age">
      </label>
```

src/scripts/main.js
```diff
@@ -1,3 +1,321 @@
...
+  });
+
+  initTableBody();
+  e.target.reset();
+  setTimeout(removeNotification, 2000);
+});
+
+tableBody.addEventListener('dblclick', (e) => {
+  const item = e.target.closest('td');
+  const i = item.cellIndex;
```
slavanosatskyi, line 265:
```suggestion
  const columnIndex = item.cellIndex;
```
name i only works in for loops, otherwise it is almost impossible to understand what does it stand for. Try to press ctrl + F and try to find in which place your variable is used ;)

slavanosatskyi (changes requested):
Very nice work, just a couple of issues to be fixed.
1. Some comments from code
2. Notification error is not clear. My form is filled, but the error says that something is missing. Try to find the way to tell the user what actually is wrong
Screenshot 2023-02-09 at 19 12 48

Volodymyr-Melnyk-Fron10der (approved):
Looks good✅

PR #304, opened 2023-02-03
solution

vladyslav-krainyk:
[DEMO LINK](https://vladyslav-krainyk.github.io/js_employees_table_DOM/)


DarkMistyRoom (changes requested):
- image

- There should be the same validation on editing as on adding
image

qoa11a (approved):

PR #307, opened 2023-02-07
table

anastasia-radchencko:
https://anastasia-radchencko.github.io/js_employees_table_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,263 @@
 'use strict';
 
-// write code here
+const table = document.querySelector('table');
+const tbody = document.querySelector('tbody');
+const tr = tbody.querySelectorAll('tr');
+const row = [...tr];
+
+// sort by click
+
+table.addEventListener('click', (e) => {
+  const th = e.target.closest('th');
+
+  if (!th || !table.contains(th)) {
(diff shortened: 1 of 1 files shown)


DarkMistyRoom (changes requested):
- the sorting works only in one direction
- all form fields should be required. I just add an empty row
image

- it appears after doubleclick
image

- there should be validation on editing
image

- use select to edit the position

slavanosatskyi (changes requested):
Screenshot 2023-02-07 at 19 00 55
Screenshot 2023-02-07 at 19 01 06
Nice work, but few more changes are needed. 
1. Please, add validation to the form. It is still possible to add name with one letter and age can be less than 18 and bigger than 90
2. On cell double click the text is not cleared, only the last word. If you change the focus previous value is not returned (see screenshots)

DarkMistyRoom (changes requested):
- Sorry, I make a mistake last time. I meant to add select on editing the office, not the position. Now I can input invalid data

image

- Salary still works incorrect
https://www.loom.com/share/9d435de147014fca91dd8ee9d99bd72c

SanyaBratashchuk (approved):

PR #309, opened 2023-02-14
Solution

Ukrainiane-panda:
[DEMO LINK](https://Ukrainiane-panda.github.io/js_employees_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,137 @@
 'use strict';
 
-// write code here
+const thead = document.querySelector('thead');
+const tbody = document.querySelector('tbody');
+
+document.body.insertAdjacentHTML('beforeend', `
+  <form action="/" method="get" class="new-employee-form">
+    <label>
+      Name:
+      <input
+        name="name"
+        type="text"
+        data-qa="name"
+        required
+        minlength="4"
+      >
+    </label>
+    <label>
+      Position:
+      <input name="position" type="text" data-qa="position" required>
+    </label>
+    <label for="office">
+      Office:
+      <select id="office" data-qa="office" required>
+        <option value="Tokyo">Tokyo</option>
+        <option value="Singapore">Singapore</option>
+        <option value="London">London</option>
+        <option value="New York">New York</option>
+        <option value="Edinburgh">Edinburgh</option>
+        <option value="San Francisco">San Francisco</option>
+      </select>
+    </label>
+    <label>
+      Age:
+      <input name="age" type="number" data-qa="age" required min="18" max="90">
+    </label>
+    <label>
+      Salary:
+      <input name="salary" type="number" data-qa="salary" required>
+    </label>
+    <button type="submit"> Save to table </button>
+  </form>
+`);
+
+const form = document.forms[0];
+const select = form.querySelector('[data-qa="office"]');
+
+form.name.addEventListener('submit', e => {
+  if (!form.name.validity.valid) {
+    form.name.setCustomValidity(
+      'Error! Less 4 letters are entered in the field'
+    );
+  }
+});
+
+form.age.addEventListener('submit', e => {
+  if (!form.age.validity.valid) {
+    form.age.setCustomValidity(
+      'Error! Age range should be from 18 to 90 years'
+    );
+  }
+});
+
+const createRow = (...arg) => {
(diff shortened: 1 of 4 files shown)


DarkMistyRoom (changes requested):
- When a user clicks on a row, it should become selected
- I can add the raw with white spaces in the name and position fields
image

- Also I can add a 3 letter name
- there is no notification element on invalid data
- "Enter" or blur don't submit the edited field

- there is a possibility to edit several cells at one time
- I can't edit the salary

ivbutsykin (changes requested):
Have some problems:

> If Name value has less than 4 letters, show error notification.
> If Age value is less than 18 or more than 90 show error notification.
1. You should show error notification like this:
Screenshot 2023-02-15 at 11 48 40

> If a new employee is successfully added to the table, show success notification.
2. Doesn't show this notification

> On 'Enter' keypress, save changes to the table cell. Remove input and set new text in the table cell.
> If an input is empty on submitting return initial value.
3. Not implemented

src/scripts/main.js
```diff
@@ -1,3 +1,220 @@
...
+      <td>${firstField}</td>
+      <td>${secondField}</td>
+      <td>${thirdField}</td>
+      <td>${fourthField}</td>
+      <td>$${format}</td>
+    </tr>
+  `);
+};
+
+function showNotifi(type, text) {
```
ivbutsykin, line 73: 😊
```suggestion
function showNotification(type, text) {
```

ivbutsykin (changes requested):
Yeah, I see that my 2 and 3 points are work fine now. 
But unfortunately error notification (1 point should be rework). Not your error show on `blur` event. Please show message only when I submit form and also I can suggest delete validation rules in your html, like `required min="18" max="90"` because it's not allow submit form.

Screenshot 2023-02-17 at 12 36 19

src/scripts/main.js
```diff
@@ -1,3 +1,227 @@
...
+
+  const notification = document.querySelector('.notification');
+
+  if (type === 'success') {
+    notification.classList.add('success');
+  };
+
+  if (type === 'error') {
+    notification.classList.add('error');
+  }
```
Anastasiia-Svintsova, line 89: Couldn't you just add type to the classList? 
```suggestion
    notification.classList.add(type);
```

Anastasiia-Svintsova (approved):
I approve it, but there are still issues that can be fixed

- I can add employees with empty fields

https://user-images.githubusercontent.com/99403198/221167640-75cd9011-cc26-4fc1-934c-4c5ee5fcb4b3.mov

- Also I can write invalid data in the salary column 
image

ivbutsykin:
@Ukrainiane-panda My response after Q/A 🙃

To avoid adding employee with spaces in name/position via form you can add check `!form.name.value.trim()` in your code, so your fnc can be (don't forget to add this for name `'change'` and position fields):
```
form.name.addEventListener('submit', e => {
  if (!form.name.validity.valid || !form.name.value.trim()) {
    form.name.setCustomValidity(
      "Error! Less 4 letters are entered in the field"
    );
  }
});
```

Next problem is changes via table. You have a big problem here:
```
validScheme.name(targetInput.value)
      || validScheme.age(targetInput.value
      || validScheme.salary(targetInput.value
```
Basically, it means if your value match with one of the patter (`validScheme` conditions) when all okay and we can changes this. **You should recognise which field was changed and check on needed schema.**

To fix empty spacing where you can add this condition (BUT it's not resolve problem above):
```
    if ((validScheme.name(targetInput.value)
      || validScheme.age(targetInput.value
      || validScheme.salary(targetInput.value))
      && !!targetInput.value.trim())) {
        ...
      }
```

P.s.: I see your PR was approved so I can suggest you familiar with my comments and be careful with validation in your next tasks.

src/scripts/main.js
Ukrainiane-panda: @ivbutsykin @Anastasiia-Svintsova thank you!  I fixed everything in the code. I added "if" for event 
```
form.addEventListener('submit', (e) => {
  e.preventDefault();
  if (form.name.value.trim()) {
    createRow();
    showNotification(
      'success',
      'Successful!',
      'New employee is successfully added to the table'
    );
  }
 e.target.reset();
});
```

PR #312, opened 2023-02-15
Finish

FallenMAD:
- [DEMO LINK](https://FallenMAD.github.io/js_employees_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -116,6 +116,7 @@ function submitFormData(e) {
 
   const data = new FormData(form);
   const employeeName = data.get('name');
+  const employeePosition = data.get('position');
   const employeeAge = data.get('age');
   const employeeSalary = data.get('salary');
   const newEmployeeSalary = '$' + String(employeeSalary);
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,243 @@
...
+let selected;
+
+thead.addEventListener('click', (e) => {
+  const item = e.target.closest('th');
+
+  const cellIndex = item.cellIndex;
+
+  if (!item) {
+    return;
+  }
```
Anastasiia-Svintsova, line 16: I think will be better to switch the order here 
```suggestion
  if (!item) {
    return;
  }
  
  const cellIndex = item.cellIndex;
```

Anastasiia-Svintsova (changes requested):
- I can add an employee with empty fields. Fix it, please 

https://user-images.githubusercontent.com/99403198/219346045-06cde9cc-5f43-42fa-bf88-aedafd29cb94.mov

- Would be good to add a min value to the age input 

https://user-images.githubusercontent.com/99403198/219346211-002fdb74-d29e-46d0-b969-9077a0f6eabe.mov

src/scripts/main.js
FallenMAD: But why?🤔

Anastasiia-Svintsova (approved):
- I approve it, but I still can add an empty position. Would be good to fix it

https://user-images.githubusercontent.com/99403198/219621777-b85ad9ac-6179-44d3-978c-a613c904a07a.mov

src/scripts/main.js
Anastasiia-Svintsova: Because why do you need this const if you don't have an item at all 🤔

PR #313, opened 2023-02-15
Solution

bohdan-gavrilyk:
https://bohdan-gavrilyk.github.io/js_employees_table_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -186,13 +186,14 @@ tableRow.addEventListener('dblclick', (e) => {
   }
 
   const indexCell = e.target.cellIndex;
-  const value = e.target.parentElement.cells[e.target.cellIndex].innerHTML;
+  let value = e.target.parentElement.cells[e.target.cellIndex].innerHTML;
   const createElement = form.elements[e.target.cellIndex].cloneNode(true);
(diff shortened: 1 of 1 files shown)


bohdan-gavrilyk:
The only thing that was not implemented was when the user clicked 2 times and left an empty line.
Remembered late =(

vlad-demchuk (changes requested):
I can't use the current value to change, I only see it as placeholder and have to type new value
image

also, implements this point
image
image

slavanosatskyi (changes requested):
1. Prev comment was not fixed, I still can't edit old value, as it is printed as placeholder
2. There is some strange behaviour with inputs 

https://user-images.githubusercontent.com/6526149/219409461-e5e3b799-20dd-494b-b203-79645161e841.mov

(see video)

slavanosatskyi (changes requested):
Screenshot 2023-02-16 at 17 58 55
1. I can't save first column values for some reason
2. 
Screenshot 2023-02-16 at 17 59 54
It is possible to edit multiple fields simultaneously. According to the task it should not be possible 
Screenshot 2023-02-16 at 17 58 12

vlad-demchuk (approved):

PR #314, opened 2023-02-16
Employees_table_DOM

ruslan-yarosh:
[DEMO LINK](https://ruslan-yarosh.github.io/js_employees_table_DOM/)


DarkMistyRoom (changes requested):
- I can add the raw with white spaces in the name and position. use `trim`

image

- Good job with validation. Would be great to show notifications on editing too

SanyaBratashchuk (approved):

PR #315, opened 2023-02-17
initial solution

Maks-Tsarenko:
[DEMO LINK](https://Maks-Tsarenko.github.io/js_employees_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -228,7 +228,7 @@ tbody.addEventListener('dblclick', e => {
   const targetCell = item.cellIndex;
   const currentItem = item.textContent;
 
-  if (targetCell < 2) {
+  if (!item.querySelector('input') && targetCell < 2) {
     const input = document.createElement('input');
 
     item.textContent = '';
@@ -270,7 +270,7 @@ tbody.addEventListener('dblclick', e => {
     });
   }
 
-  if (targetCell === 2) {
+  if (!item.querySelector('input') && targetCell === 2) {
     const select = document.createElement('select');
 
     select.className = 'cell-input';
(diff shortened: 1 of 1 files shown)


SanyaBratashchuk (changes requested):
I can add employee without salary 
image

src/scripts/main.js
```diff
@@ -1,3 +1,396 @@
...
+    input.type = 'text';
+
+    item.appendChild(input);
+    input.focus();
+
+    input.addEventListener('keydown', ev => {
+      if (ev.key === 'Enter') {
+        if (input.value.length < 4) {
+          pushNotification(200, 10, 'Invalid name',
+            'The name must consist of at least 4 characters.', 'error');
```
Volodymyr-Melnyk-Fron10der, line 239: I suggest to move all messages into constants

src/scripts/main.js
```diff
@@ -1,3 +1,396 @@
...
+      <td>${age}</td>
+      <td>${salary}</td>
+     </tr>
+  `);
+
+  pushNotification(100, 10, 'The operation is successful',
+    'Employee was successfully added', 'success');
+});
+
+tbody.addEventListener('dblclick', e => {
```
Volodymyr-Melnyk-Fron10der, line 219: What about to move commands for every targetCell case into separated functions and call them?  It would be easier to read

Volodymyr-Melnyk-Fron10der (changes requested):
1. Form should be cleaned after submitting
image
2. It works with bugs if use those buttons 
image

Volodymyr-Melnyk-Fron10der (changes requested):
image
I can leave an empty value, try to fix it pls

Volodymyr-Melnyk-Fron10der (approved):
Well done💪🏼

PR #317, opened 2023-02-19
Solution

vitalii-borzov:
- [DEMO LINK](https://vitalii-borzov.github.io/js_employees_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,430 @@
 'use strict';
 
 // write code here
+// all needed variables
+
+const body = document.querySelector('body');
+const thead = document.querySelector('thead');
+const tbody = document.querySelector('tbody');
+const rows = [...tbody.rows];
+const nameLength = 4;
+const minAge = 18;
+const maxAge = 90;
+const timeoutOfMessage = 3000;
+let compareIndex = -1;
+let isValid;
+let editingArea;
+let isCorrect = true;
+
+// table sorting
+
+thead.addEventListener('click', eventClick => {
+  const th = eventClick.target.closest('th');
+  const index = th.cellIndex;
+
+  sortTableAsc(index, th.innerText, compareIndex === index);
+
+  compareIndex = (compareIndex === index) ? -1 : index;
+});
+
+function sortTableAsc(index, columnName, dscSorting) {
+  rows.sort((first, second) => {
+    const a = first.children[index].innerHTML;
+    const b = second.children[index].innerHTML;
+
+    switch (columnName) {
+      case 'Name':
+      case 'Position':
+      case 'Office':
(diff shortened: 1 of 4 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,430 @@
...
+    isValid = true;
+  }
+};
+
+addEmployee.addEventListener('submit', submitEvent => {
+  const employeeInfo = new FormData(addEmployee);
+  const salaryToNormalize = employeeInfo.get('salary');
+  let normalizedSalary = salaryToNormalize;
+
+  if (salaryToNormalize.length > 3) {
```
TarasHoliuk, line 161: It works only with the first 3 characters:
image
[toLocaleString()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toLocaleString) will be useful here

src/scripts/main.js
```diff
@@ -1,3 +1,430 @@
...
+  setTimeout(() => {
+    message.remove();
+  }, timeoutOfMessage);
+};
+
+// change, validate and save data in cell-input
+
+tbody.addEventListener('dblclick', doubleClick => {
+  const target = doubleClick.target.closest('td');
+  const input = document.querySelector('.cell-input');
```
TarasHoliuk, line 224: show the initial value in input:
image

src/scripts/main.js
```diff
@@ -1,3 +1,430 @@
...
+      }
+
+      break;
+  }
+}
+
+function editEnd(td, isOk) {
+  if (isOk) {
+    switch (td.cellIndex) {
+      case 4:
```
TarasHoliuk, line 391: editing the current field every time leads to NaN:
image
image

src/scripts/main.js
```diff
@@ -1,3 +1,430 @@
...
+          460,
+          10,
+          'Changed successfully!',
+          'Congrats! The employee data was successfully changed.',
+          'success');
+        break;
+    }
+  } else {
+    if (td.firstChild.value === '') {
+      td.innerHTML = editingArea.data;
```
TarasHoliuk, line 417: but for the first two columns you show validation error in this case

src/scripts/main.js
```diff
@@ -1,3 +1,430 @@
...
+      return;
+    }
+
+    return editStart(td);
+  }
+
+  editingArea = null;
+};
+
+function changeIsCorrect() {
```
TarasHoliuk, line 428:
```suggestion
function resetIsCorrect() {
```
since you always set the same (initial) value

TarasHoliuk (changes requested):
GJ :+1: 
I've added a few suggestions
Also, a few points should be fixed (check comments)

PR #318, opened 2023-02-19
Employees Table

Igor-Yaremchuk:
[EMPLOYEES TABLE](https://Igor-Yaremchuk.github.io/js_employees_table_DOM/)


src/scripts/main.js
```diff
@@ -1,3 +1,324 @@
...
+  }
+}
+
+/* ============== INPUTS ENTRIES VALIDATION ============== */
+
+function checkUserEntries(entries) {
+  const specification = {
+    'name': {
+      rules: (value) => value.length < 4,
+      description: 'Name must contain at least 4 letters',
```
Volodymyr-Melnyk-Fron10der, line 106: I suggest to move all success messages into SuccessMessages object and errors into ErrorsMessages

Volodymyr-Melnyk-Fron10der (changes requested):
Hi! Great idea to use classes and OOP here👍

1. Increase a width of the input 
image

polosanya (approved):

PR #320, opened 2023-02-20
js for sorting and changing data in the table and form were added

Olena-Shevchenko:
- [DEMO LINK](https://Olena-Shevchenko.github.io/js_employees_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,335 @@
 'use strict';
 
-// write code here
+function convertSalary(salary) {
+  return +salary.slice(1).split(',').join('');
+};
+
+function convertNumberToSalary(number) {
+  const symbol = '$';
+  const digit = new Intl.NumberFormat('en-US').format(number);
+
+  return symbol + digit;
+}
+
+function sortTableString(columnNumber, inputOrder,
+  inputLastSorted, columnName) {
+  let ascOrder = inputOrder;
+
+  if (inputLastSorted !== columnName) {
+    ascOrder = true;
+  };
+
+  if (ascOrder) {
+    rows.sort((a, b) => a.children[columnNumber].textContent
+      .localeCompare(b.children[columnNumber].textContent));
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,335 @@
...
-// write code here
+function convertSalary(salary) {
+  return +salary.slice(1).split(',').join('');
+};
+
+function convertNumberToSalary(number) {
+  const symbol = '$';
+  const digit = new Intl.NumberFormat('en-US').format(number);
+
+  return symbol + digit;
```
witflash, line 11: I prefer to use string interpolation for such cases: 
```
return `$${digit}`
```
But it's ok )

src/scripts/main.js
```diff
@@ -1,3 +1,335 @@
...
+}
+
+function sortTableString(columnNumber, inputOrder,
+  inputLastSorted, columnName) {
```
witflash, line 15:
```suggestion
function sortTableString(
  columnNumber, 
  inputOrder,
  inputLastSorted, 
  columnName
) {
```

witflash: It's better to place each argument on a new line if you can't place in a single line.

src/scripts/main.js
```diff
@@ -1,3 +1,335 @@
...
+  notification.insertAdjacentHTML('afterbegin', `
+    <h2 class="title">${title}</h2>
+    <p>${message}</p>
+  `);
+
+  body.append(notification);
+
+  setTimeout(() => {
+    notification.remove();
+  }, '3000');
```
witflash, line 190: You can place such a code in single line:
```
setTimeout(() => notification.remove(), 3000);
```
But it's not critical. If you like your syntax, it's ok.
And it's better to use a `number` type for the timeout.

witflash (approved):
Good work!
I left some uncritical comments for your future work.

PR #323, opened 2023-02-22
Solution

pushkarskiyrodion:
[DEMO LINK](https://pushkarskiyrodion.github.io/js_employees_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,242 @@
 'use strict';
 
-// write code here
+const table = document.querySelector('table');
+const tBody = document.querySelector('tbody');
+
+table.addEventListener('click', (e) => {
+  const employees = document.querySelector('tbody').querySelectorAll('tr');
+
+  if (e.target.tagName === 'TD') {
+    const hasActive = [...employees]
+      .find(element => element.classList.contains('active'));
+
+    if (hasActive) {
+      hasActive.removeAttribute('class');
+    }
+
+    const tr = e.target.closest('tr');
+
+    tr.classList.add('active');
+  }
+
+  if (e.target.tagName === 'TH') {
+    sortGrid(e.target.cellIndex, [...employees]);
+  }
+});
+
+let wasSorted = false;
+
+function sortGrid(index, array) {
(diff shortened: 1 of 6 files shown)


SanyaBratashchuk (changes requested):
Can't see new employee in the table

pushkarskiyrodion:
The new employee will add at the end of the table. Doesn't work?

src/scripts/main.js
```diff
@@ -1,3 +1,322 @@
...
+    array.sort((a, b) => {
+      const first = a.cells[index].innerText.replace(/[$,]/g, '');
+      const second = b.cells[index].innerText.replace(/[$,]/g, '');
+
+      if (isNaN(first)) {
+        return second.localeCompare(first);
+      }
+
+      return second - first;
+    });
```
ivbutsykin, line 64: It looks like very similar part of code. Can we create separate function to do this?

src/scripts/main.js
```diff
@@ -1,3 +1,322 @@
...
+};
+
+tBody.addEventListener('dblclick', (e) => {
+  const value = e.target.innerText;
+  const input = document.createElement('input');
+  const td = e.target.closest('td');
+
+  input.style.width = '90px';
+  input.classList.add('cell-input');
+  input.style.width = '90px';
```
ivbutsykin, line 182: We set `input.style.width` two times?

src/scripts/main.js
```diff
@@ -1,3 +1,322 @@
...
+
+  if (input.name === 'salary') {
+    input.value = input.value > 0
+      ? input.value
+      : pushNotification(10, 10, 'Error message',
+        'Enter correct data', 'error', input);
+  }
+
+  return input.value;
+}
```
ivbutsykin, line 286: Will be nice create constants for validation numbers and input names:
```
// Choose name instead FOO =)
const FOO =  {
  NAME: 'name',
  AGE: 'age',
  ...
};
const  MAX_NAME_LENGTH  = 4;
...

ivbutsykin (changes requested):

PR #333, opened 2023-03-05
Develop

IrynaK-a:
[DEMO LINK](https://IrynaK-a.github.io/js_employees_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -6,13 +6,106 @@ const tHeaders = Array.from(document.body.querySelector('thead')
   .children
 );
 const tBody = document.body.querySelector('tbody');
-const tBodyRows = Array.from(tBody.children);
+let tBodyRows = Array.from(tBody.children);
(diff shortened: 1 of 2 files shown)


polosanya (changes requested):
1. It would be great to specify the error text, so user knows what exactly is wrong
image
2. Edited value is not saved on "Enter"
image
image

src/scripts/main.js
```diff
@@ -1,3 +1,367 @@
...
+          newInputValue = newInput.value;
+      }
+    }
+
+    target.textContent = (newInput.value === '') ? targetValue : newInputValue;
+
+    newInput.remove();
+  });
+
+  newInput.addEventListener('keydown', (ev) => {
```
anstsot, line 339: you have same event listener on line 305. so you can merge them

anstsot (approved):
Great job, approving

PR #339, opened 2023-03-08
Ready

sergik94:
https://sergik94.github.io/js_employees_table_DOM/


src/scripts/main.js
```diff
@@ -1,3 +1,249 @@
...
+function setOrderTrue(arr) {
+  arr.forEach(th => {
+    th.dataset.ascOrder = true;
+  });
+}
+
+function sortList(e) {
+  const columnIndex = e.target.cellIndex;
+  const currCol = e.target;
+  const falseList = thead.querySelectorAll('[data-asc-order="false"]');
```
slavanosatskyi, line 75: not the best name. It can be misleading. try to come up with more domain related name like ascendingSortedColumns or something like this

src/scripts/main.js
```diff
@@ -1,3 +1,249 @@
...
+  let sortedArr;
+
+  if (currCol.dataset.ascOrder === 'true') {
+    setOrderTrue([...falseList]);
+
+    currCol.dataset.ascOrder = false;
+
+    sortedArr = [...employeeList].sort((employee1, employee2) => {
+      const data1 = formatSalary(employee1, columnIndex);
+      const data2 = formatSalary(employee2, columnIndex);
```
slavanosatskyi, line 85:
```suggestion
      const currentSalary = formatSalary(employee1, columnIndex);
      const nextSalary = formatSalary(employee2, columnIndex);
```
employee1 and employee2 can be renamed as well

src/scripts/main.js
```diff
@@ -1,3 +1,249 @@
...
+
+      if (Number(data1)) {
+        return data1 - data2;
+      }
+
+      return data1.localeCompare(data2);
+    });
+  } else {
+    setOrderTrue([...falseList]);
```
slavanosatskyi, line 94: you perform this command under if statement as well. Can you just move it outside of "if" statement

src/scripts/main.js
```diff
@@ -1,3 +1,249 @@
...
+
+      return data2.localeCompare(data1);
+    });
+  }
+
+  tbody.append(...sortedArr);
+}
+
+function selectRow(e) {
+  const isActive = tbody.querySelector('.active');
```
slavanosatskyi, line 112: isActive sounds like a boolean variable. In fact it is an element. Please, rename it correspondingly - something like activeRow would work

src/scripts/main.js
```diff
@@ -1,3 +1,249 @@
...
+    isActive.classList.remove('active');
+
+    return;
+  }
+
+  if (isActive) {
+    isActive.classList.remove('active');
+  }
+
+  targetRow.classList.add('active');
```
slavanosatskyi, line 125:
```suggestion
  if (![...targetRow.classList].includes('active')) {
    targetRow.classList.add('active');
  }

  if (isActive) {
    isActive.classList.remove('active');
  }
```

slavanosatskyi (changes requested):
Seems that sorting doesn't count newly added rows
Screenshot 2023-03-08 at 18 04 30

src/scripts/main.js
sergik94: @slavanosatskyi , NO. Performing of this function depends on the CURRENT "data-order-state" of column. And, if I  move it outside of "if" statement, "data-order-state" will always be "true"

src/scripts/main.js
sergik94: @slavanosatskyi  oooohh...  I thought that we must have opportunity to deactivate current raw.

etojeDenys (approved):
looks good

PR #341, opened 2023-03-09
https://nastychenko.github.io/js_employees_table_DOM/

nastychenko:


qoa11a (changes requested):
Please provide demo link.

nastychenko:
> Almost done! Let's also fix the error notifications. Increase the z-index of notifications so that they appear on top of the form image

It's not a z-index problem, it's a transparent background problem.
I can change it if needed

sTorba24 (changes requested):
Good work! Small thing to be fixed
when i add new employee it's not participating in filtering

src/scripts/main.js
```diff
@@ -1,3 +1,310 @@
...
+    input.className = 'cell-input';
+    editCell.textContent = '';
+    editCell.append(input);
+    input.focus();
+
+    const saveInputValue = e => {
+      if (e.key === 'Enter' || e.type === 'blur') {
+        editCell.textContent = input.value;
+
+        if (input.value.trim() === '') {
```
sTorba24, line 297:
```suggestion
        if (!(input.value.trim())) {
```

anstsot (changes requested):
Looks good but there are 2 issues to fix:
1. the newly added person isn't sorted by clicking on table header
2. it is possible to fill chars instead of numbers in Age and Salary
image

sTorba24 (changes requested):
You did great correction of mistakes 👍 
Lets finish with them

shouldn't be possible to set name and position using numbers
image

agniya-i (approved):

PR #342, opened 2023-03-09
solution

MSA1991:
- [DEMO LINK](https://msa1991.github.io/js_employees_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,243 @@
 'use strict';
 
-// write code here
+const list = document.querySelector('tbody');
+let sortReverse = false;
+
+const head = document.querySelector('thead').children[0];
+const headList = [...head.querySelectorAll('th')];
+
+// -----------------SORT LIST------------------------
+
+head.addEventListener('click', (e) => {
+  const item = e.target;
+  const indexItem = headList.indexOf(item);
+  const person = [...list.querySelectorAll('tr')];
+
+  sortList(person, indexItem);
+});
+
+function sortList(arr, i) {
+  if (!sortReverse) {
+    sortReverse = true;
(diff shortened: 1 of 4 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,243 @@
...
+    }
+  }
+
+  if (isFieldsFilled) {
+    pushNotification('error', 'Fill in all the fields', 'error');
+
+    return;
+  } else {
+    pushNotification('success', 'New person added', 'success');
+  }
```
Volodymyr-Melnyk-Fron10der, line 156: you don't need an `else` if you do return in the `if` block
```suggestion
  } 
  
  pushNotification('success', 'New person added', 'success');
  
```

src/scripts/main.js
```diff
@@ -1,3 +1,243 @@
...
+
+  field.focus();
+
+  field.addEventListener('blur', () => {
+    if (field.value.length === 0) {
+      item.innerHTML = itemValue;
+    } else {
+      field.dataset.qa === 'salary'
+        ? item.innerHTML = getSalary(field.value)
+        : item.innerHTML = field.value;
```
Volodymyr-Melnyk-Fron10der, line 201: maybe it's better to do it only for salary field?

Volodymyr-Melnyk-Fron10der (changes requested):
Hi, good job! Check some comments pls. And great file structure, by the way👍

1. looks like the value is not saved when I use arrows
image
image

MSA1991:
Что касается стрелок оно все сохраняет, просто в вашем случае число 4 не проходит валидацию.

mykhalenych (approved):

PR #346, opened 2023-03-11
Solution

vlad-ptashnichenko:
[DEMO LINK](https://vlad-ptashnichenko.github.io/js_employees_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,309 @@
 'use strict';
 
-// write code here
+const thead = document.querySelector('thead');
+const headTableRow = thead.querySelector('tr');
+const tbody = document.querySelector('tbody');
+const body = document.querySelector('body');
+
+let countClick = 0;
+const enterKeyCode = 13;
+let initialText = null;
+let selectedCell = null;
+
+const form = document.createElement('form');
+const inputForm = document.createElement('input');
+const inputName = ['name', 'position', 'office', 'age', 'salary'];
+const optionName = [`Tokyo`, `Singapore`, `London`, `New
+York`, `Edinburgh`, `San Francisco`];
+const labelText = ['Name:', 'Position:', 'Office:', 'Age:', 'Salary:'];
+const label = document.createElement('label');
+const select = document.createElement('select');
+const option = document.createElement('option');
+const button = document.createElement('button');
+
+for (let i = 0; i < labelText.length; i++) {
+  label.textContent = labelText[i];
+  form.append(label.cloneNode(true));
+}
+
+for (let i = 0; i < optionName.length; i++) {
+  option.value = optionName[i];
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,309 @@
...
+        + inputValues[inputValue].slice(1, inputValues[inputValue].length);
+        newTr.append(td.cloneNode(true));
+      } else if (inputValue === 'salary') {
+        td.textContent = toCashString(inputValues[inputValue].toString());
+        newTr.append(td.cloneNode(true));
+      } else {
+        td.textContent = inputValues[inputValue];
+        newTr.append(td.cloneNode(true));
+      }
+    }
```
VitaliyBondarenko1982, line 97: try to avoid else if.
if this case suggesting to replace `if ... else if ... else` on `switch...case`.

src/scripts/main.js
```diff
@@ -1,3 +1,309 @@
...
+      notification(
+        'Error',
+        'The age value should be between 18 and 90 years',
+        'error'
+      );
+
+      return initialText;
+    } else {
+      return inputValue;
+    }
```
VitaliyBondarenko1982, line 197:
```suggestion
    } 
    
     if (inputValue > 90 || inputValue < 18) {
      notification(
        'Error',
        'The age value should be between 18 and 90 years',
        'error'
      );

      return initialText;
    }
      
      return inputValue;
```

src/scripts/main.js
```diff
@@ -1,3 +1,309 @@
...
+  }
+
+  if (columnIndex === 4) {
+    if (valueHasLetters(inputValue)) {
+      return initialText;
+    } else {
+      return toCashString(inputValue);
+    }
```
VitaliyBondarenko1982, line 205:
```suggestion
    } 
    
    return toCashString(inputValue);
```

src/scripts/main.js
```diff
@@ -1,3 +1,309 @@
...
+        'Error',
+        'A word cannot start with numbers',
+        'error'
+      );
+
+      return initialText;
+    } else {
+      return inputValue.charAt(0).toUpperCase()
+        + inputValue.slice(1, inputValue.length);
+    }
```
VitaliyBondarenko1982, line 181:
```suggestion
    if (('0123456789').includes(inputValue.charAt(0))) {
      notification(
        'Error',
        'A word cannot start with numbers',
        'error'
      );

      return initialText;
    } 
     return inputValue.charAt(0).toUpperCase()
        + inputValue.slice(1, inputValue.length);
```

VitaliyBondarenko1982 (changes requested):
Good job.
Suggest some refactoring.

etojeDenys (approved):
good job

PR #347, opened 2023-03-15
Task done

andriy-litkovets:
[DEMO LINK](https://andriy-litkovets.github.io/js_employees_table_DOM/)


qoa11a (changes requested):
- Clear the inputs after the form submission.
- When I add data to the table and sort it - the added data is removed. Fix it.
- The user shouldn't be able to add an employee with a negative salary.
- I can edit the age to be negative or less than 18.

etojeDenys (approved):
LGTM

PR #350, opened 2023-03-18
First solution

Ssviatt:
- [DEMO LINK](https://Ssviatt.github.io/js_employees_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,266 @@
 'use strict';
 
-// write code here
+const tableBlock = document.querySelector('table');
+
+const getNumbers = (num) => {
+  return num.replaceAll(',', '').replace('$', '');
+};
+const formatter = new Intl.NumberFormat('en-US', {
+  style: 'currency',
+  currency: 'USD',
+  maximumFractionDigits: 0,
+});
+
+const dataType = (value) => isNaN(+getNumbers(value)) ? 'string' : 'number';
+
+const pushNotification = (posTop, posRight, title, description, type) => {
+  document.body.insertAdjacentHTML('beforeend', `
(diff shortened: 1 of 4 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,266 @@
...
+const tableBlock = document.querySelector('table');
+
+const getNumbers = (num) => {
+  return num.replaceAll(',', '').replace('$', '');
+};
+const formatter = new Intl.NumberFormat('en-US', {
+  style: 'currency',
+  currency: 'USD',
+  maximumFractionDigits: 0,
+});
```
ivbutsykin, line 12: Nice decision to use it!

src/scripts/main.js
```diff
@@ -1,3 +1,266 @@
...
+    e.preventDefault();
+
+    const newName = form.elements.name.value;
+    const position = form.elements.position.value;
+    const office = form.elements.office.value;
+    const age = form.elements.age.value;
+    const salary = form.elements.salary.value;
+
+    switch (true) {
+      case String(newName).length < 4:
```
ivbutsykin, line 139: I can suggest move `4, 18, 90` and other magic numbers into constants

ivbutsykin (changes requested):
Almost done, resolve bottom issues, please.
1. Age and salary inputs should have a number type (allow user type only numbers):
Screenshot 2023-03-18 at 13 54 22

2. User can add employee with empty name (if I add 4 spaces instead letters):
Screenshot 2023-03-18 at 13 59 52

3. User can make salary as `NaN` (to do this input letters when edit employee salary):
Screenshot 2023-03-18 at 14 01 44

SanyaBratashchuk (approved):

PR #352, opened 2023-03-22
Solution

undepend:
[DEMO LINK](https://undepend.github.io/js_employees_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,296 @@
 'use strict';
 
-// write code here
+const thead = document.querySelector('thead');
+const headers = thead.querySelectorAll('th');
+const tbody = document.querySelector('tbody');
+
+// Table sorting by clicking on the title (in two directions)
+
+thead.addEventListener('click', function(e) {
+  const targetIndex = e.target.cellIndex;
+  const sort = [...tbody.rows].sort((a, b) => {
+    let cellA = a.cells[targetIndex].innerHTML;
+    let cellB = b.cells[targetIndex].innerHTML;
+
+    if (cellA[0] === '$' || cellB[0] === '$') {
+      cellA = +cellA.slice(1).replace(',', '');
+      cellB = +cellB.slice(1).replace(',', '');
+
+      return cellA - cellB;
+    }
+
+    if (isFinite(cellA) || isFinite(cellB)) {
+      return cellA - cellB;
+    }
+
+    return cellA.localeCompare(cellB);
+  });
+
+  if (e.target.classList.contains('asc')) {
+    e.target.classList.remove('asc');
+    tbody.append(...sort.reverse());
+  } else {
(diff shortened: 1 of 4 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,296 @@
...
+  } else if (dataObject.age < 18 || dataObject.age > 90) {
+    pushNotification('error', 'Error', 'Age must be between 18 to 90');
+  } else if (!dataObject.position || dataObject.position.length < 2) {
+    pushNotification('error', 'Error', 'Enter valid position');
+  } else if (!dataObject.salary || isNaN(dataObject.salary)) {
+    pushNotification('error', 'Error', 'Enter valid salary');
+  } else {
+    pushNotification('success', 'Success', 'New employee was added');
+    withoutMistakes = true;
+  }
```
Viktor-Kost, line 139: You can use sth like this to avoid else if nesting (it's hard to read)

``` 
const validations = {
    name: {
      isValid: dataObject.name.length >= 4,
      message: 'Name could not be less than 4 letters'
    },
    age: {
      isValid: dataObject.age >= 18 && dataObject.age <= 90,
      message: 'Age must be between 18 to 90'
    },
    position: {
      isValid: dataObject.position && dataObject.position.length >= 2,
      message: 'Enter valid position'
    },
    salary: {
      isValid: !isNaN(dataObject.salary) && dataObject.salary > 0,
      message: 'Enter valid salary'
    }
  };

  Object.keys(validations).forEach(key => {
    const validation = validations[key];

    if (!validation.isValid) {
      pushNotification('error', 'Error', validation.message);
      withoutMistakes = false;
    }
  })
  ```

File: src/scripts/main.js [hunk 2]
```diff
@@ -1,3 +1,296 @@
...
+        if (input.value.length < 4) {
+          pushNotification('error', 'Error',
+            'Name should have more than 4 letters');
+          e.target.textContent = prevText;
+
+          return;
+        } else {
+          e.target.textContent = input.value;
+        }
+        break;
```
Viktor-Kost, line 213: why do you need both `return` and `break` ?

src/scripts/main.js
```diff
@@ -1,3 +1,296 @@
...
+
+      case 2:
+        if (input.value) {
+          e.target.textContent = input.value;
+        } else {
+          e.target.textContent = prevText;
+        }
+        break;
```
Viktor-Kost, line 232: use `ternary operator` for such cases

Viktor-Kost (changes requested):
Hi!)

Please, fix this behavior for salary (it should be like toggler. First click from the biggest to the lows, second click - opposite behavior)

https://user-images.githubusercontent.com/117587463/227004447-befa221a-2267-4ba7-a4ae-4c794a12a2d2.mov

etojeDenys (approved):
lgtm

PR #355, opened 2023-03-24
Solution

8Melya8:
- [DEMO LINK](https://8melya8.github.io/js_employees_table_DOM/)


VitaliyBondarenko1982 (changes requested):
Godd job.
Let's fix notification.
They should be as on previous task
Something like that
Screenshot from 2023-03-24 20-57-22

if enter several times on submit button they looks little weird
Screenshot from 2023-03-24 20-47-12

etojeDenys (changes requested):
almost done
let's reset the form after adding a new item to table

VitaliyBondarenko1982 (approved):
Good job.

PR #356, opened 2023-03-28
JS code for employees table and new employe form

ilona-romaniuk:
https://ilona-romaniuk.github.io/js_employees_table_DOM/


qoa11a (changes requested):
image
- It's still possible to add an employee with an empty name.

anstsot (changes requested):
Good work!
Found one more issue with fields editing. I'm changing the age but it remains the same. Salary works as expected.
image
image

PR #363, opened 2023-04-02
Solution

LBondarchuk:
- [DEMO LINK](https://LBondarchuk.github.io/js_employees_table_DOM/)

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,4 +1,3 @@
 module.exports = {
-  extends: ['@mate-academy/eslint-config', 'plugin:cypress/recommended', 'plugin:chai-friendly/recommended'],
-  plugins: ['cypress', 'chai-friendly'],
+  extends: ['@mate-academy/eslint-config', 'plugin:cypress/recommended'],
 };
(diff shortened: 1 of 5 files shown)


LBondarchuk:
Hello! according to the tests, it gives one error. Perhaps it is not suitable for tests, but the code works

LBondarchuk:
AssertionError: Timed out retrying: expected '<div.notification.success>' to have class 'error'
    at Context.eval (http://localhost:8080/__cypress/tests?p=cypress/integration/jsEmployeesTableDOM.spec.js:337:40)
cy.fillNewEmployeeForm('Ada', 'QA Engineer', 'San Francisco', 18, 50000);
cy.get('[data-qa="notification"]').should('have.class', 'error');
cy.checkDataDoesntExist('Ada', 50000);

LBondarchuk:
but here I see that the tests did not pass at all.

polosanya (changes requested):
Good job! Let's improve your solution a bit)
1. It would be great to specify what exactly user need to enter. Also, all notifications should disappear after a few seconds
image
2. It is possible to add a user with empty fields if add spaces instead of name and position
image
image
3. It would be great to add ability to select the offices instead of editing manually
image

qoa11a (changes requested):
- Don't clear inputs if there's an error.
- The user shouldn't be able to enter an empty name/position/age/salary when editing existing data in the table.
- The user shouldn't be able to enter non-numeric values inside the salary and age fields when editing. 
- The user shouldn't be able to enter an age less than 18 or more than 90 when editing age.

sTorba24 (approved):
Well done!

PR #365, opened 2023-04-06
First commit

dudar-y:
[DEMO LINK](https://dudar-y.github.io/js_employees_table_DOM/)

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,4 +1,3 @@
 module.exports = {
-  extends: ['@mate-academy/eslint-config', 'plugin:cypress/recommended', 'plugin:chai-friendly/recommended'],
-  plugins: ['cypress', 'chai-friendly'],
+  extends: ['@mate-academy/eslint-config', 'plugin:cypress/recommended'],
 };
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,358 @@
 'use strict';
 
-// write code here
+const tableBody = document.querySelector('tbody');
+const tableHead = document.querySelector('thead');
+const rows = [...tableBody.rows];
+
+// Sorting table
+
+let typeOfSort = 'DESC';
+let sortPrevTarget;
+
+function sortTable(index, array, type) {
+  const text = rows[0].cells[index].innerText;
+
+  const sortedRow = type === 'ASC'
+    ? array.sort((rowA, rowB) => {
+      if (!isNaN(Number(text))) {
+        return rowA.cells[index].innerText - rowB.cells[index].innerText;
+      };
+
+      if (text[0] === '$') {
+        return convertToNumber(rowA.cells[index].innerText)
+          - convertToNumber(rowB.cells[index].innerText);
+      };
+
+      return rowA.cells[index].innerText
+        .localeCompare(rowB.cells[index].innerText);
+    })
+    : array.sort((rowB, rowA) => {
+      if (!isNaN(Number(text))) {
+        return rowA.cells[index].innerText - rowB.cells[index].innerText;
+      };
+
+      if (text[0] === '$') {
+        return convertToNumber(rowA.cells[index].innerText)
+        - convertToNumber(rowB.cells[index].innerText);
+      };
+
+      return rowA.cells[index].innerText
+        .localeCompare(rowB.cells[index].innerText);
+    });
+
+  tableBody.append(...sortedRow);
+}
+
+tableHead.addEventListener('click', (e) => {
+  const index = [...tableHead.rows[0].cells].indexOf(e.target);
+
+  if (typeOfSort === 'DESC' || sortPrevTarget !== e.target) {
+    sortTable(index, [...tableBody.rows], 'ASC');
+    typeOfSort = 'ASC';
+  } else {
+    sortTable(index, [...tableBody.rows], 'DESC');
+    typeOfSort = 'DESC';
+  };
+
+  sortPrevTarget = e.target;
+});
+
+function convertToNumber(string) {
+  return +string.slice(1).split(',').join('');
+};
+
+// Selected Row
+
+let selectedElement;
+
+tableBody.addEventListener('click', (e) => {
+  const row = e.target.parentElement;
+
+  if (!selectedElement) {
+    row.classList.add('active');
+    selectedElement = row;
+  } else if (selectedElement !== row) {
+    selectedElement.classList.remove('active');
+
+    row.classList.add('active');
+    selectedElement = row;
+  };
+});
+
+// AddForm Script
+
+const cityesOfOffice = [
+  'Tokyo',
+  'Singapore',
+  'London',
+  'New York',
+  'Edinburgh',
+  'San Francisco',
+];
+
+// eslint-disable-next-line no-shadow
+function createInput(name, type) {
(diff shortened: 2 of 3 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,358 @@
...
+// Sorting table
+
+let typeOfSort = 'DESC';
+let sortPrevTarget;
+
+function sortTable(index, array, type) {
+  const text = rows[0].cells[index].innerText;
+
+  const sortedRow = type === 'ASC'
+    ? array.sort((rowA, rowB) => {
```
SerhiiKirik, line 16: DRY - don't repeat yourselft
Move this check ( type === 'ASC' ) - to the sort callback

SerhiiKirik: also, split to the variables

src/scripts/main.js
```diff
@@ -1,3 +1,358 @@
...
+  e.valid = true;
+
+  if (inputName.firstElementChild.value.length < 4) {
+    pushNotification(10, 10, 'Name input error',
+      'The Name field\n '
+    + 'contain more than three letters.', 'error');
+    e.valid = false;
+  };
+
+  if (inputAge.firstElementChild.value < 18
```
SerhiiKirik, line 215: split to the variable

SerhiiKirik (changes requested):
Well done 🚀 
Let's improve your code & solution 🧑‍💻 

mb would be better to start notifications from the upside screen?
image

src/scripts/main.js
```diff
@@ -1,3 +1,360 @@
...
+
+  const nameValue = inputName.firstElementChild.value;
+  const ageValue = inputAge.firstElementChild.value;
+  const positionValue = inputPosition.firstElementChild.value;
+  const salaryValue = inputSalary.firstElementChild.value;
+
+  if (nameValue.length < 4) {
+    pushNotification(10, 10, 'Name input error',
+      'The Name field\n '
+    + 'contain more than three letters.', 'error');
```
ivanchuk-yurii, line 206:
```suggestion
    pushNotification(
      10,
      10,
      'Name input error',
      'The Name field contain more than three letters.',
      'error',
    );
```

ivanchuk-yurii: fix in other places

src/scripts/main.js
```diff
@@ -1,3 +1,360 @@
...
+const cityesOfOffice = [
+  'Tokyo',
+  'Singapore',
+  'London',
+  'New York',
+  'Edinburgh',
+  'San Francisco',
+];
+
+// eslint-disable-next-line no-shadow
```
ivanchuk-yurii, line 83: don't disable rules, fix errors instead

src/scripts/main.js
```diff
@@ -1,3 +1,360 @@
...
+  };
+
+  setTimeout(removeNotification, 3000);
+};
+
+// Form validation
+
+buttonSumbit.addEventListener('click', e => {
+  e.preventDefault();
+  e.valid = true;
```
ivanchuk-yurii, line 196: why do you change event? I think it's better to create variable

src/scripts/main.js
```diff
@@ -1,3 +1,360 @@
...
+    } else {
+      cell.innerText = '';
+    }
+
+    cell.append(cellInput);
+    cellInput.focus();
+  }
+});
+
+document.addEventListener('click', (e) => {
```
ivanchuk-yurii, line 314: use blur event for input instead of document click and editableCell variable

ivanchuk-yurii (changes requested):
add validation also for inserted fields in table, not just in form. when field appears in table it should have default value of cell. office field should be select, not input

src/scripts/main.js
```diff
@@ -1,3 +1,424 @@
...
+    selectedElement = row;
+  } else if (selectedElement !== row) {
+    selectedElement.classList.remove('active');
+
+    row.classList.add('active');
+    selectedElement = row;
+  };
+});
+
+// AddForm Script
```
DarkMistyRoom, line 72: ok, but for the future, think about using `insertAdjacentHTML` in a similar situation. it may be shorter

src/scripts/main.js
```diff
@@ -1,3 +1,424 @@
...
+  document.body.append(notification);
+
+  const removeNotification = () => {
+    notification.remove();
+  };
+
+  setTimeout(removeNotification, 3000);
+};
+
+// Form validation
```
DarkMistyRoom, line 197: it is ok, but in the future, you can make validation logic shorter by using `FormData` object. Read about it.

DarkMistyRoom (changes requested):
Great job! Few moments:

- remove the previous message before showing the next one
image

- fix the possibility to add employ with whitespaces in the name and position fileds
image

- pay attention to this requirement
image

.eslintrc.js
```diff
@@ -1,4 +1,3 @@
 module.exports = {
-  extends: ['@mate-academy/eslint-config', 'plugin:cypress/recommended', 'plugin:chai-friendly/recommended'],
-  plugins: ['cypress', 'chai-friendly'],
```
SerhiiKirik, line 3: The 'plugin:chai-friendly/recommended' configuration is missing. Please add it back to the 'extends' array to ensure adherence to the recommended code style linting. 🧐

.eslintrc.js
```diff
@@ -1,4 +1,3 @@
...
+  extends: ['@mate-academy/eslint-config', 'plugin:cypress/recommended'],
```
SerhiiKirik, line 2: The 'chai-friendly' plugin is missing from the 'plugins' array. Please add it back to ensure adherence to the recommended code style linting. 😊

src/scripts/main.js
```diff
@@ -1,3 +1,438 @@
...
+  } else {
+    sortTable(indexForSorting, [...tableBody.rows], 'DESC');
+    typeOfSort = 'DESC';
+  };
+
+  sortPrevTarget = e.target;
+});
+
+function convertToNumber(string) {
+  return +string.slice(1).split(',').join('');
```
SerhiiKirik, line 51: Consider using Number() instead of + for type conversion.

src/scripts/main.js
```diff
@@ -1,3 +1,438 @@
...
+    const fieldsOfForm = [...document.querySelectorAll('label')];
+    const employeeData = {};
+
+    fieldsOfForm.forEach(field => {
+      const input = field.firstElementChild;
+
+      employeeData[input.name] = input.value;
+    });
+
+    employeeData.salary = '$' + Number(employeeData.salary)
```
SerhiiKirik, line 269: Use template literals instead of string concatenation for better readability.

src/scripts/main.js
```diff
@@ -1,3 +1,438 @@
...
+// Sorting table
+
+let typeOfSort = 'DESC';
+let sortPrevTarget;
+
+function sortTable(sortIndex, array, type) {
+  const text = rows[0].cells[sortIndex].innerText;
+
+  const getSortedRow = (a, b) => {
+    if (!isNaN(Number(text))) {
```
SerhiiKirik, line 16: Use Number.isNaN() instead of !isNaN() for better readability.

SerhiiKirik (approved):
Well done 🔥

PR #366, opened 2023-04-06
Solution

nazarkharkevych:
https://nazarkharkevych.github.io/js_employees_table_DOM/

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 1. Replace `<your_account>` with your Github username in the link
-    - [DEMO LINK](https://<your_account>.github.io/js_employees_table_DOM/)
+    - [DEMO LINK](https://nazarkharkevych.github.io/js_employees_table_DOM/)
 2. Follow [this instructions](https://mate-academy.github.io/layout_task-guideline/)
     - Run `npm run test` command to test your code;
     - Run `npm run test:only -- -n` to run fast test ignoring linter;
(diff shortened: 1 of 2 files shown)


slavanosatskyi (changes requested):
It is possible to save empty value in a table. The value should be restored to previous in this case.
Table is getting wider during cell editing 
https://user-images.githubusercontent.com/6526149/230425959-73a46869-af0e-4fc4-8127-df5302ab3be9.mov

SanyaBratashchuk (changes requested):
I can add employee with name which contains only spaces. Also when I enter 1 in salary field it turn into 0,001$ 
image

SerhiiKirik (changes requested):
Well done 🚀 
is something wrong with the form

https://user-images.githubusercontent.com/91411692/230563484-db140d82-d94c-4ed1-9f9f-a71bbe41fdc6.mov

Also, i can update the age with words
image

loralevitska (changes requested):
Keep up the good work! Some comments to fix:
- limit the age to the maximum value, for example 100 years or as in the form and the salary field validate on numbers
image
- clear form after submit 
image
- make salary value only positive number

Volodymyr-Melnyk-Fron10der (changes requested):
there should be dropdown as well as in the form
image

SanyaBratashchuk (approved):

PR #367, opened 2023-04-06
Solution

nadia-hope:
[DEMO LINK](https://nadia-hope.github.io/js_employees_table_DOM/)


qoa11a (changes requested):
- Don't clear inputs if there was an error.
- Make select of possible cities when editing a city in an existing row.
- Currently, I can edit a name or a position and make it empty (if I just enter space).
- I can enter letters when editing salary.

anstsot (changes requested):
Looks good but there are few issues to fix:
1. I can add an employee with an empty name and position
image

2. it would be better to add more informative errors to the form
3. I can change age to less than 18 and more than 99
image

agniya-i (approved):

PR #372, opened 2023-04-11
Develop

Vaskovics:
<a href='https://vaskovics.github.io/js_employees_table_DOM/'>Demo Link</a>

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,7 +1,5 @@
 'use strict';
 
-const log = console.log;
-
 // Sorting:
 const tableHead = document.querySelector('thead');
 const allRows = [...document.querySelector('table').rows];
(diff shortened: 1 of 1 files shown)


DarkMistyRoom (changes requested):
Good, but let's improve!
- Please pay attention to this requirement

image

- enter the added salary in the correct format
- please specify the error messages. there is no information about the required fields and requirements to input data, but always the same message
image

- I can add employ with only white spaces in the name
image

- I can input invalid data by editing
image

- 
image

polosanya (changes requested):
Almost done! A few more fixes left 🙂  
There is an error only for age, despite the fact that it's correct. Add errors for other fields too
image
Also, it is still possible to add a user with empty fields (by adding spaces for name input)
image

Vaskovics:
Thank you! seems like I didn't test it properly.
Also if you could give me some feedback about my code, is it readable? or maybe implantation is not the best?

SerhiiKirik (changes requested):
Great job 🔥 
Let's fix some bugs 🧑‍💻 
I can add & update a salary to less than 0
image

Vaskovics:
Basically it should be alright, as soon junior developers will have to pay to be hired :)))

Volodymyr-Melnyk-Fron10der (changes requested):
1. salary should be required as well. And you should reset the form after submitting
image

2. there should be a select component
image

OleziO (approved):

src/scripts/main.js
```diff
@@ -1,3 +1,337 @@
...
+      tableDataToSort.sort((a, b) => {
+        const toNumberA = a.Salary.match(/\d+/g).join('');
+        const toNumberB = b.Salary.match(/\d+/g).join('');
+
+        return toNumberA - toNumberB;
+      });
+    } else {
+      tableDataToSort.sort((a, b) =>
+        a[sortByColumn].localeCompare(b[sortByColumn]));
+    }
```
DarkMistyRoom, line 56: looks like duplication. I can suggest another approach. split callback choosing and order choosing logic. Use variable for a callback, set it based on clicked column, and then set the order flag, and in the end use them for sorting. Fell free to ask if you have questions

DarkMistyRoom (changes requested):
- as for me would be better if a current option were selected on double-clicking on the office field, but it's up to you

- check the salary editing. It always gives an error. I think that because there is a string after transformation but you expect a number

DarkMistyRoom (approved):
Looks good to me)

PR #373, opened 2023-04-11
Employees table was improved

serge-kontakt:
- [DEMO LINK](https://serge-kontakt.github.io/js_employees_table_DOM/)

There is just one problem. If we start sort the table every other features do not work then. Can you give an advice how to fix it?

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,291 @@
 'use strict';
 
 // write code here
+
+function convertToNumber(str) {
+  const numb = +str.replace(/\D/g, '');
+
+  return numb;
+}
+
+let clickNumber0 = 0;
+let clickNumber1 = 0;
+let clickNumber2 = 0;
+let clickNumber3 = 0;
+let clickNumber4 = 0;
+
+const sortAll = (e) => {
+  switch (e.target) {
+    case document.querySelectorAll('th')[0]:
+      clickNumber0 += 1;
+      clickNumber1 = 0;
+      clickNumber2 = 0;
+      clickNumber3 = 0;
+      clickNumber4 = 0;
+      break;
+
+    case document.querySelectorAll('th')[1]:
+      clickNumber0 = 0;
+      clickNumber1 += 1;
+      clickNumber2 = 0;
+      clickNumber3 = 0;
+      clickNumber4 = 0;
+      break;
+
+    case document.querySelectorAll('th')[2]:
+      clickNumber0 = 0;
+      clickNumber1 = 0;
+      clickNumber2 += 1;
+      clickNumber3 = 0;
+      clickNumber4 = 0;
+      break;
+
+    case document.querySelectorAll('th')[3]:
+      clickNumber0 = 0;
+      clickNumber1 = 0;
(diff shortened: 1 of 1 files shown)


qoa11a (changes requested):
What do you mean by saying "If we start sort the table every other features do not work then."? Please, ask questions in fe_chat and continue our communication there.

image
- Validate fields when editing a cell. Currently, I can set age and salary to strings, set empty values etc. Apply the same validation you have for the form to add a new employee for editing cells.

- Trim inputs when adding a new employee. Currently, I can add an employee with an empty name if I just put spaces inside of the input.

qoa11a (changes requested):
Good job! 👍 

https://user-images.githubusercontent.com/102293862/231547597-9bd0bd32-4dea-4936-8151-c725292839e8.mov

- Fix this issue.
- Make select for cities when editing this cell.

src/scripts/main.js
```diff
@@ -1,3 +1,324 @@
...
+      clickNumber3 = 0;
+      clickNumber4 = 0;
+      break;
+
+    case document.querySelectorAll('th')[3]:
+      clickNumber0 = 0;
+      clickNumber1 = 0;
+      clickNumber2 = 0;
+      clickNumber3 += 1;
+      clickNumber4 = 0;
```
OleziO, line 60: If you will use an array, you don't need all these checks.

src/scripts/main.js
```diff
@@ -1,3 +1,324 @@
...
+<option value="London">London</option>
+<option value="New York">New York</option>
+<option value="Edinburgh">Edinburgh</option>
+<option value="San Francisco">San Francisco</option>
+</select></label>
+<label>Age: <input name="age" type="number" data-qa="age" required></label>
+<label>Salary: <input name="salary" type="number" data-qa="salary" required>
+</label>
+<button type="submit">Save to table</button>
+`;
```
OleziO, line 176: Format this code. Add spaces before each nested line.

src/scripts/main.js
```diff
@@ -1,3 +1,324 @@
...
+
+formElement.addEventListener('submit', (e) => {
+  e.preventDefault();
+
+  const data = new FormData(formElement);
+
+  if (data.get('name').trim().length < 4
+  || data.get('age') < 18
+  || data.get('age') > 90
+  || data.get('position').trim() === '') {
```
OleziO, line 195: Add two spaces before each line.

src/scripts/main.js
```diff
@@ -1,3 +1,324 @@
...
+
+    const salaryNumber = +newEmployee.salary;
+
+    newEmployee.salary = '$' + salaryNumber.toLocaleString('en-US');
+
+    const newRow = document.createElement('tr');
+
+    for (const key in newEmployee) {
+      const newCell = document.createElement('td');
+
```
OleziO, line 208: Remove unnecessary empty lines. Check other similar cases.

OleziO (changes requested):
Make error text more meaningful. Add min and max values for number inputs.
image

PR #377, opened 2023-04-15
Done

Ilia991:
https://Ilia991.github.io/js_employees_table_DOM/


OleziO (changes requested):
image
Add field validation.

anstsot (changes requested):
The validation seems to work but there are no error notifications. could you please add them? Because I was a bit confused why the fields become empty and there are no notifications to say what's wrong. 
image
image

PR #378, opened 2023-04-16
solution

marynatovtyn:
[DEMO LINK](https://marynatovtyn.github.io/js_employees_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,267 @@
 'use strict';
 
-// write code here
+const body = document.querySelector('body');
+const table = document.querySelector('table');
+const tableBody = table.querySelector('tbody');
+const thead = table.querySelector('thead');
+const tableHeaders = thead.querySelectorAll('th');
+
+let sortOrder = 1;
+let activeHeader = null;
+const offices = [`Tokyo`, `Singapore`, `London`,
+  `New York`, `Edinburgh`, `San Francisco`];
+
+// Create form
+const form = document.createElement('form');
+const saveBtn = document.createElement('button');
+
+form.classList.add('new-employee-form');
+
+tableHeaders.forEach(tableHeader => {
+  const label = document.createElement('label');
+  const input = document.createElement('input');
+
+  label.textContent = tableHeader.textContent;
+
+  if (tableHeader.textContent === 'Office') {
+    const select = createSelect(tableHeader);
+
+    label.append(select);
+    form.append(label);
+
+    return;
+  }
+
+  tableHeader.textContent === 'Age' || tableHeader.textContent === 'Salary'
+    ? input.type = 'number'
+    : input.type = 'text';
+
+  input.name = tableHeader.textContent.toLowerCase();
+  input.required = true;
+  input.setAttribute('data-qa', tableHeader.textContent.toLowerCase());
+
+  label.append(input);
+  form.append(label);
+});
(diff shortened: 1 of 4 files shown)


SanyaBratashchuk (changes requested):
image
Validation on edit should be the same as on create

src/scripts/main.js
```diff
@@ -1,3 +1,300 @@
...
+const thead = table.querySelector('thead');
+const tableHeaders = thead.querySelectorAll('th');
+
+let sortOrder = 1;
+let activeHeader = null;
+let currentCell = false;
+const offices = [`Tokyo`, `Singapore`, `London`,
+  `New York`, `Edinburgh`, `San Francisco`];
+
+// Create form
```
OleziO, line 15: Remove all comments.

OleziO (changes requested):
image
Add validation to all fields.

src/scripts/main.js
```diff
@@ -1,3 +1,300 @@
...
+const table = document.querySelector('table');
+const tableBody = table.querySelector('tbody');
+const thead = table.querySelector('thead');
+const tableHeaders = thead.querySelectorAll('th');
+
+let sortOrder = 1;
+let activeHeader = null;
+let currentCell = false;
+const offices = [`Tokyo`, `Singapore`, `London`,
+  `New York`, `Edinburgh`, `San Francisco`];
```
DarkMistyRoom, line 13:
```suggestion
const offices = [
  `Tokyo`, 
  `Singapore`, 
  `London`,
  `New York`, 
  `Edinburgh`, 
  `San Francisco`
];
```

src/scripts/main.js
```diff
@@ -1,3 +1,300 @@
...
+
+    label.append(select);
+    form.append(label);
+
+    return;
+  }
+
+  tableHeader.textContent === 'Age' || tableHeader.textContent === 'Salary'
+    ? input.type = 'number'
+    : input.type = 'text';
```
DarkMistyRoom, line 37: try to avoid combining ternary with other logical operators

DarkMistyRoom: use variables

src/scripts/main.js
```diff
@@ -1,3 +1,300 @@
...
+  form.append(label);
+});
+saveBtn.textContent = 'Save to table';
+form.append(saveBtn);
+body.append(form);
+
+function normaliseNumber(num) {
+  return parseFloat(num.replace('$', '').replace(',', ''));
+};
+editTableCell(table);
```
DarkMistyRoom, line 53:
```suggestion
});

saveBtn.textContent = 'Save to table';
form.append(saveBtn);
body.append(form);

function normaliseNumber(num) {
  return parseFloat(num.replace('$', '').replace(',', ''));
};

editTableCell(table);
```

DarkMistyRoom (changes requested):
- the notification says that all fields should have at least 4 characters, but is it applied to age and salary?

image

- I can add employ with only white spaces in name and position fields
image

- reset the form after submit
- please specify which age is valid
image

- If an input is empty on submitting return initial value.(editing)

SanyaBratashchuk (changes requested):
image
When I start editing the default value of input should be value of current cell

SanyaBratashchuk (changes requested):
image
image

OleziO (approved):

PR #380, opened 2023-04-17
solution

AnastassiaH:
[DEMO LINK](https://anastassiah.github.io/js_employees_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -17,13 +17,14 @@ const sortTable = (sortBy, order) => {
 
     valueB = !order ? [valueA, valueA = valueB][0] : valueB;
 
-    if (+valueA) {
+    if (Number(valueA)) {
       return valueA - valueB;
     }
 
-    if (+valueA.slice(1).split(',').join('')) {
-      return +valueA.slice(1).split(',').join('')
-      - +valueB.slice(1).split(',').join('');
+    const isSalary = Number(valueA.slice(1).split(',').join(''));
+
+    if (isSalary) {
+      return isSalary - Number(valueB.slice(1).split(',').join(''));
     }
 
     return valueA > valueB ? 1 : valueA < valueB ? -1 : 0;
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,189 @@
...
+    let valueA = a.children[index].textContent;
+    let valueB = b.children[index].textContent;
+
+    valueB = !order ? [valueA, valueA = valueB][0] : valueB;
+
+    if (+valueA) {
+      return valueA - valueB;
+    }
+
+    if (+valueA.slice(1).split(',').join('')) {
```
Viktor-Kost, line 24: Instead of using `+` to convert a string to a number, use `Number()` or `parseInt()` with a radix of 10. And it's better to create a variable for this condition

src/scripts/main.js
```diff
@@ -1,3 +1,189 @@
...
+<button type="submit">Save to table</button>
+`);
+document.body.append(newEmployeeForm);
+
+// validate function
+
+const validateForm = function(elements) {
+  elements.pop();
+
+  switch (true) {
```
Viktor-Kost, line 92: I would recommend to refactor this validation function:
1) it's better not use switch statement like this. You can use just early return
Example: 
```
  const isEmpty = rest.some(item => !item.value.length);
  if (isEmpty) {
    pushNotification('Error', 'All fields are required', 'error');
    return false;
  }
```
2) Use a named constant for the minimum name length, minimum and maximum age (outside the function)
```
  const MIN_NAME_LENGTH = 4;
  const MIN_AGE = 18;
  const MAX_AGE = 90;
  ```
  3) just create a variables for conditions
  Example:
  ```
  const isNameTooShort = rest[0].value.length < MIN_NAME_LENGTH;
  ```

Viktor-Kost (changes requested):
Good job!
Some moments can be improved

YegorKorenkov (approved):
Looks good

PR #383, opened 2023-04-19
Develop

Grechana:
https://Grechana.github.io/js_employees_table_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -130,24 +130,13 @@ button.addEventListener('click', (e) => {
 
   const salaryFormal = '$' + Intl.NumberFormat('en-US').format(salary);
 
-  if (person.length < 4) {
+  if (person.length < 4 || position.length < 4) {
     body.append(div);
     div.dataset.qa = `notification`;
     div.className = `notification error`;
 
     div.innerHTML = `<h2 class='title'>ERROR </h2>
-    <p> Name  has less than 4 letters</p>`;
-
-    return;
-  }
-
-  if (position.length < 4) {
-    body.append(div);
-    div.dataset.qa = `notification`;
-    div.className = `notification error`;
-
-    div.innerHTML = `<h2 class='title'>ERROR </h2>
-    <p> position  has less than 4 letters</p>`;
+    <p>  Text has less than 4 letters</p>`;
 
     return;
   }
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,220 @@
 'use strict';
 
-// write code here
+const tableHeader = document.querySelector('thead');
+
+const rowList = document.querySelectorAll('tbody >tr');
+const tbody = document.querySelector('tbody');
+
+const thead = document.querySelector('thead>tr');
+const headerNames = document.querySelectorAll('thead>tr>th');
+
+for (const nameClicked of headerNames) {
+  nameClicked.isClicked = false;
+};
+
+tableHeader.addEventListener('click', (e) => {
+  const item = e.target.textContent;
+
+  let children = 0;
+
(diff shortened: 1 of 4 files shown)


polosanya (changes requested):
Good job! Let's improve your solution a bit)
1. It is possible to create a user with empty name and position (just add some spaces instead of them)
image
2. It would be great for notifications to disappear after a few seconds
image
3. It is possible to set incorrect values on editing
image
4. It would be great to implement ability to save the changes on Enter
image

vlad-demchuk (changes requested):
I can't change office filed. it doesn't same my new value
image

the notification is displayed only the first time. also, can we not reset all fields if some data is wrong?
https://user-images.githubusercontent.com/102411130/233567912-bb35cfc4-dce2-47f3-98bd-da02143c3b20.mov

src/scripts/main.js
```diff
@@ -1,3 +1,294 @@
...
+    }
+  }
+
+  const listArr = [...rowList];
+
+  if (e.target.isClicked === false) {
+    if (item === 'Salary') {
+      listArr.sort((x, y) =>
+        (+y.children[children].textContent.replace(/,/g, '').replace('$', ''))
+      - (+x.children[children].textContent.replace(/,/g, '').replace('$', '')));
```
Anastasiia-Svintsova, line 32: You write the same logic twice. Move it to a separate function

src/scripts/main.js
```diff
@@ -1,3 +1,294 @@
...
+
+  for (let i = 0; i < thead.children.length; i++) {
+    if (item === thead.children[i].textContent) {
+      children = i;
+    }
+  }
+
+  const listArr = [...rowList];
+
+  if (e.target.isClicked === false) {
```
Anastasiia-Svintsova, line 28: You can use a logical operator instead
```suggestion
  if (!e.target.isClicked) {
```

src/scripts/main.js
```diff
@@ -1,3 +1,294 @@
...
+
+          notificationError();
+        }
+      }
+      editedCell = null;
+    });
+
+    input.addEventListener('keydown', (ev) => {
+      if (ev.key === 'Enter') {
+        if ((cell === cell.parentElement.children[0])
```
Anastasiia-Svintsova, line 254: You can use switch/case construction to avoid these if statements

Anastasiia-Svintsova (changes requested):
- The sorting doesn't work for added values 

https://user-images.githubusercontent.com/99403198/233609862-6e854644-8bc1-4440-95c2-be9fe2bd1406.mov

- I can add an invalid value to the office field. Show the input while editing there 

https://user-images.githubusercontent.com/99403198/233610372-304ce202-53c4-4e37-bc14-fb1edb7b3553.mov

agniya-i (changes requested):
Good job! Let's fix one more bug)
After I change Office, the dropdown doesn't disappear

https://user-images.githubusercontent.com/32853034/233714143-0aa885f8-f521-4bc1-94a0-bace926009af.mov

ivanchuk-yurii (changes requested):
still not fixed. try to redeploy your changes
image

OleziO (changes requested):
image
Make error text more meaningful.
image
Add validation to the salary field.

ivanchuk-yurii (changes requested):
after dbclick on office value and focusout `select` doesn't disappear 
image
after adding first employee, it's possible to add second without office and salary values
image
fix this point
image

polosanya (approved):
Well done! Just found one small bug.
There is a wrong error message when trying to edit the office for incorrect value. It would be great to select the office o
image
image

PR #384, opened 2023-04-20
solution

DanielSorokowski:
- [DEMO LINK](https://DanielSorokowski.github.io/js_employees_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,200 @@
 'use strict';
 
-// write code here
+const tableHead = document.querySelector('thead');
+const tableBody = document.querySelector('tbody');
+const elements = tableBody.querySelectorAll('tr');
+const users = [];
+
+tableHead.addEventListener('click', e => {
+  const element = e.target;
+
+  if (element.tagName === 'TH') {
+    if (!element.getAttribute('data-sort')) {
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,200 @@
...
+    });
+
+    element.classList.add('active');
+  });
+});
+
+const createForm = () => {
+  const newForm = document.createElement('form');
+
+  newForm.classList.add('new-employee-form');
```
d-wasilewski, line 107: Instead of disabling eslint `max-len`, consider breaking the long string into multiple lines, this shouldn't break anything.

src/scripts/main.js
```diff
@@ -1,3 +1,200 @@
...
+  document.body.appendChild(notification);
+
+  setTimeout(() => {
+    document.body.removeChild(notification);
+  }, 2000);
+};
+
+let currentValue;
+
+tableBody.addEventListener('click', (clickEvent) => {
```
d-wasilewski, line 173: Consider using event delegation for handling the editing of table cells, instead of adding event listeners for each cell.

d-wasilewski (changes requested):
[Optional]
- Only one input should be edited at a time
image
- cells should become editable on double click, instead of single clikc

d-wasilewski (approved):

PR #386, opened 2023-04-23
solution

Nikoramo:
- [DEMO LINK](https://Nikoramo.github.io/js_employees_table_DOM/)


DarkMistyRoom (changes requested):
- a comma should be inserted after every 3 numbers. just use `.toLocaleString()`

image

- normalize inputted values. names that started with whitespace go before others
image

- 
image

PR #387, opened 2023-04-24
add task solution

meltourme:
https://meltourme.github.io/js_employees_table_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,315 @@
 'use strict';
 
-// write code here
+const thead = document.querySelector('thead');
+const headTableRow = thead.querySelector('tr');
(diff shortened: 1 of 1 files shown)


meltourme:
I've already done it. Please check it again or explain me what I have to do)

> use defined values for Office field as in form image image

 I've already done it. Please check it again or explain me what I have to do)

loralevitska (changes requested):
Apply all comments both to the form and to the edit line in the table:

- Should be select by edit row 
image
- Also make this fields required 
image
- The salary value must be greater than zero 
image

If you have any questions, please ask them in fe_chat and tag mentors. It will be faster, because your answers in GitHub will be visible to mentor only after you re-request review

PR #388, opened 2023-04-25
Solution

natalia-maryshko:
[DEMO LINK](https://natalia-maryshko.github.io/js_employees_table_DOM/)


OleziO (changes requested):
Add string validation.
image

polosanya (approved):
Good job!
Why does the notification appear here and not in the top right corner, like in previous tasks?
You can change z-index of the notification
image

PR #389, opened 2023-04-26
solution part of the task

NadieinOleh:
https://NadieinOleh.github.io/js_employees_table_DOM/


agniya-i (changes requested):
1. When a validation error appears and then I try to add the correct data, the error text is still presented even if it's a success message 
2. Sort does not work after I add the user 
3. It would be better to reset the form after adding a new user 
4. Add more descriptive error messages
5. Add user to the end of the list

https://user-images.githubusercontent.com/32853034/234660233-f179118f-00c9-481d-a1f4-d92c35595a0b.mov

vlad-demchuk (approved):

PR #392, opened 2023-05-02
Employees table solution v.1

annperetiatko:
[DEMO LINK]
https://annperetiatko.github.io/js_employees_table_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,271 @@
 'use strict';
 
 // write code here
+
+// task 1 - sort table //
+const table = document.querySelector('table');
+const listOfRows = document.querySelectorAll('tbody tr');
+const columnHeaders = table.querySelectorAll('thead th');
+const rows = table.querySelector('tbody');
+let sortCounter = 0;
+
+columnHeaders.forEach((header, i) => {
+  header.addEventListener('click', () => {
+    const rowsArray = Array.from(rows.querySelectorAll('tr'));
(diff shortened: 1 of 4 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,271 @@
 'use strict';
 
 // write code here
```
OleziO, line 3: Remove all comments.

OleziO (changes requested):
Add salary validation
image

OleziO (changes requested):
Not resolved.
image

anstsot (changes requested):
Good job but there are few issues.
1. It's possible to put the non-numeric characters. 
image

2. The strange behaviour of inputs:

https://user-images.githubusercontent.com/22824431/235910782-1d1aec50-1588-4bca-ad0d-958fd2d03bb5.mov

3. It's possible to change to empty value
image

tetiana-ivanchenko (changes requested):
if I add salary 5000 it's appear in the table as 5000000

tetiana-ivanchenko (changes requested):
still the same problem if I add 20

anstsot (changes requested):
Good progress but I've found 2 more issues:
1. The newly added employee can't be changed in the table 
2. the Office can be changed to other values (maybe it would be better to use select?)
image

OleziO (approved):

PR #393, opened 2023-05-03
Table sorting by clicking on the title was implemented

oksanatytanych:
- [DEMO LINK](https://oksanatitanych.github.io/js_employees_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,39 @@
 'use strict';
 
-// write code here
+const header = document.querySelectorAll('thead tr th');
+
+[...header].map((element, index) => {
+  element.addEventListener('click', (e) => {
+    const rows = document.querySelectorAll('tbody tr');
+
+    if (!element.hasAttribute('direction')
+      || element.getAttribute('direction') === 'DESC') {
+      element.setAttribute('direction', 'ASC');
+    } else {
+      element.setAttribute('direction', 'DESC');
+    }
+
+    const sortedRows = [...rows].sort((a, b) => {
+      let compareA = a.querySelectorAll('td')[index].innerText;
+      let compareB = b.querySelectorAll('td')[index].innerText;
+
+      if (element.getAttribute('direction') === 'DESC') {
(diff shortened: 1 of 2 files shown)


oksanatytanych:
Table sorting by clicking on the title was implemented

src/scripts/main.js
```diff
@@ -1,3 +1,39 @@
 'use strict';
 
-// write code here
+const header = document.querySelectorAll('thead tr th');
+
+[...header].map((element, index) => {
```
YegorKorenkov, line 5: Why you use map method? You don't need to return a new array, please use foreach

YegorKorenkov (changes requested):
Looks good, please complete the remaining tasks
image

oksanatytanych:
> * [DEMO LINK](https://oksanatytanych.github.io/js_employees_table_DOM/)

src/scripts/main.js
```diff
@@ -1,3 +1,222 @@
...
+      fAge.focus();
+
+      return false;
+    } else {
+      pushNotification('New employee is successfully', 'success');
+
+      return true;
+    }
+  }
+}
```
IvanFesenko, line 160: 2023-05-15_15h58_06

I can add only spaces

Try refactor without using else statement
```
function valideData() {
  if (newEmployy.includes('')) {
    pushNotification('All fields must be filled', 'warning');
    return false
  } 
    const fName = form.querySelector('[name="name"]');
    const fAge = form.querySelector('[name="age"]');

    if (fName.value.length < 4) {
      pushNotification('Name minimum length is 4', 'error');
      fName.focus();

      return false;
    } 
    .....
}
```

IvanFesenko (changes requested):
Also please move notification a bit bottom, cuz for now it's block form
2023-05-15_16h01_19

etojeDenys (changes requested):
let's create select when users try to edit this field
image

2. let's validate all inputs using also `.trim()` to avoid redundant spaces
image

PR #394, opened 2023-05-05
The table of employees has been improved

kateryna-stemblivska:
[DEMO LINK](https://kateryna-stemblivska.github.io/js_employees_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -73,7 +73,9 @@ employeeForm.addEventListener('submit', (e) => {
   const data = new FormData(employeeForm);
 
   if (data.get('name').trim().length < 4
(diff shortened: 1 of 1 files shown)


loralevitska (changes requested):
Good job! There are things that should be fixed:

- Add validate for `name` and `position` fields. Now I added spaces
image
- add select for `office` column for edit option like in form
image
- Validate these fields
image

anstsot (changes requested):
Good job, but it's still possible to add empty data (just with spaces)
image

witflash (changes requested):
Check my last comment. It may help you fix the problem with fields that contain spaces only. 

image

PR #397, opened 2023-05-09
solution

danulyk05:


sTorba24 (changes requested):
Provide demo link to review it

danulyk05:
https://danulyk05.github.io/js_employees_table_DOM/

SanyaBratashchuk (changes requested):
1. It would be better to specify which field doesn't pass validation 
image
2. Validation on edit should be the same as on create 
image

src/scripts/main.js
```diff
@@ -1,3 +1,339 @@
...
+
+  [...tableRow].map(tr => {
+    [...tr.children].forEach((td, index) => {
+      td.dataset.nameIs = arrHead[index];
+    });
+  });
+}
+
+let rows;
+// ---------------------------------------------------------
```
OleziO, line 21: Remove all comments from the code

src/scripts/main.js
```diff
@@ -1,3 +1,339 @@
...
+
+    input.name = th.textContent.toLowerCase();
+    input.dataset.qa = th.textContent.toLowerCase();
+
+    switch (th.textContent) {
+      case 'Age':
+      case 'Salary':
+        input.type = 'number';
+        break;
+
```
OleziO, line 71: Remove the empty line

OleziO (changes requested):

SanyaBratashchuk (approved):

PR #398, opened 2023-05-10
solution

mariuszgit:


Radoslaw-Czerniawski (changes requested):
Generally there are a few problems:
- on row modification I can add a string to number fields and numbers to string fields
image
- fields that I've added manually are not editable
image

- I've also managed to execute an xss attack on your pr by pasting this into the input on the form

```html
<img src="x" onerror="document.querySelector('table')?.remove(); document.body.innerHTML = '<span>haha zostałeś zhakowany</span>'"> 
```

image

- the last bit is caused by setting innerHTML on the row, the image tag from the script I showed you before is pasted into your markup and becomes an image with invalid src causing the code on onerror to be exectued, this can be prevented by explicitly creating html elements, adding them textContent and then appending them

Radoslaw-Czerniawski (changes requested):
- Newly added fields don't take part in filtering
- I can input numbers into text inputs for newly created fields
image

mariuszgit:
Is it possible that Youre not using latest ver of my commits? :D
Everyuthing should work fine.

pon., 15 maj 2023 o 18:23 Radosław Czerniawski ***@***.***>
napisał(a):

Radoslaw-Czerniawski (approved):
Generally some things could be shortened, some things could be exported to different files but I won't bother you with that, overall the implementation works, and seems bulletproof so good job :D

PR #400, opened 2023-05-14
Solution

Ihor-Zakharov:
[DEMO LINK](https://ihor-zakharov.github.io/js_employees_table_DOM/)


SanyaBratashchuk (changes requested):
1. Improve validation. I can add negative salary
image
2. Salary is not formatted after adding

SanyaBratashchuk (changes requested):
1. 
image
2. 
image
3. Validation on edit should be the same as on add 
image

PR #403, opened 2023-05-18
Solution

HannaVasylieva:
https://HannaVasylieva.github.io/js_employees_table_DOM/


OleziO (changes requested):
The notification should not shift the layout:
image

Add fields validation before adding:
image

PR #405, opened 2023-05-20
Solution

yura-enttsel:
[DEMO LINK](https://yura-enttsel.github.io/js_employees_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,306 @@
 'use strict';
 
-// write code here
+const table = document.querySelector('table');
+const tableBody = table.querySelector('tbody');
+const tableHeader = table.querySelector('thead');
+
+// Implementation of table sorting by clicking on the title (in two directions)
+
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,306 @@
...
+        >
+      `;
+      break;
+
+    default:
+      selectedCell.innerHTML = `
+        <input
+          class="cell-input"
+          type="text"
+          pattern="[a-zA-Z0-9]+[a-zA-Z0-9 ]+"{4}
```
vlad-demchuk, line 239: Remove the '{4}' from the pattern attribute in the input element, as it doesn't seem to have any effect and may cause confusion. Use minLength attribute instead.

vlad-demchuk (changes requested):
specify what exactly is wrong
image

nazarmatsevych (approved):
👍

PR #407, opened 2023-05-22
add task solution

Ir-ra:
[DEMO LINK](https://Ir-ra.github.io/js_employees_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,228 @@
 'use strict';
 
-// write code here
+const head = document.querySelector('thead');
+const tbody = document.querySelector('tbody');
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,228 @@
 'use strict';
 
-// write code here
+const head = document.querySelector('thead');
+const tbody = document.querySelector('tbody');
+const body = document.querySelector('body');
+
+// -----sorting employees-----
```
witflash, line 7: You can remove any comments

witflash (approved):
Well done 👍 
There is no need to add the comments you added. Your code is self-describable.

PR #415, opened 2023-05-27
add task solution

valeriaoleynik:
[DEMO LINK](https://valeriaoleynik.github.io/js_employees_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,280 @@
 'use strict';
 
-// write code here
+const table = document.querySelector('table');
+const tbody = table.querySelector('tbody');
+const thead = document.querySelectorAll('thead th');
+
+const getNumber = (value) => {
+  return Number(value.slice(1).split(',').join(''));
+};
+const convertSalary = (value) => {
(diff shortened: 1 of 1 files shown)


Anastasiia-Svintsova (changes requested):
- I can add an employee with empty fields. Fix it, please 

https://github.com/mate-academy/js_employees_table_DOM/assets/99403198/aa3ee469-b005-4d83-b951-5ec0b01d60c2

- The user should be able to write only valid values to the office field while editing. You can use a selector inside the table as well
image

- The user is not able to edit added employees for now. Fix it, please 

- Would be great to add min attributes to these fields 
image

qoa11a (changes requested):
image
Validate all the fields on edit, please.

lerastarynets (changes requested):
All good except for notifications) I think you should be able to see Error notifications as shown in the beginning of video, however on validation of all fields except Name a default html notification is shown. Using `event.preventDefault` may help you with this problem)

https://github.com/mate-academy/js_employees_table_DOM/assets/77551820/4409edfd-767a-4d47-82a2-73b9eee4a6c8

PR #425, opened 2023-06-02
add task solution 1

denys2:
Your site is live at https://denys2.github.io/js_employees_table_DOM/


witflash (changes requested):
Good job!
One thing.
Could you implement the next behavior? If a user starts field editing you have to pass the existing value as an initial value, but not an empty field.
image

denys2:
check please

agniya-i (approved):

PR #426, opened 2023-06-06
table done

Alessandranitti15:
[DEMO LINK](https://Alessandranitti15.github.io/js_employees_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -254,7 +254,7 @@ document.body.appendChild(form);
 // changing cell text
 
 tableBody.addEventListener('dblclick', (e) => {
-  const input = document.createElement('input');
+  let input = document.createElement('input');
   const td = e.target;
   const cellIndex = td.cellIndex;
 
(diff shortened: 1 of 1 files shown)


qoa11a (changes requested):
Here's the review of your work: https://drive.google.com/file/d/1unywK79R9zUrLj_YQCUG4dDK6i6Znxjn/view?usp=sharing

src/scripts/main.js
```diff
@@ -1,3 +1,306 @@
...
+  if (person.length < 4) {
+    pushNotification(450, 10, 'Error!',
+      'Name must have minimum 4 lettes', 'error');
+
+    return;
+  }
+
+  if (position.length === 0) {
+    pushNotification(450, 10, 'Error!',
+      'Name must have minimum 4 lettes', 'error');
```
lerastarynets, line 203: Update error message to 'Position must have at least 1 character' since it's related to position length.

src/scripts/main.js
```diff
@@ -1,3 +1,306 @@
 'use strict';
 
-// write code here
+const tableHead = document.querySelector('thead');
+const tableBody = document.querySelector('tbody');
+const rows = document.querySelectorAll('tr');
```
lerastarynets, line 5: Consider using `tableBody.querySelectorAll('tr')` instead of `document.querySelectorAll('tr')` to select only rows within the table body.

lerastarynets (changes requested):
Great job overall)) However, you need to fix error messages and there is an issue that one can set whichever value to the Office field while editing

https://github.com/mate-academy/js_employees_table_DOM/assets/77551820/9cb23384-24be-4c55-b7e4-56b6c013d65d

lerastarynets (changes requested):
Well done, now your notifications are correct, however i can still type any value into Office field while editing) Try adding same select to the row in editing mode with options from which user can choose

polosanya (approved):
Good job!

PR #430, opened 2023-06-08
message

Pa1eOrc:
[DEMO LINK](https://Pa1eOrc.github.io/js_employees_table_DOM/)


DarkMistyRoom (changes requested):
- Please pay attention to this requirement
image

- I can add employee with only white spaces in the name and position. fix it
image

- the added employee doesn't take part in the sorting

- add validation on editing. and use `select` for office editing

PR #431, opened 2023-06-09
add task solution

antonina-klishch:
https://antonina-klishch.github.io/js_employees_table_DOM/


Volodymyr-Melnyk-Fron10der (changes requested):
1. should be select 
image

PR #433, opened 2023-06-10
solution

Gosyav:
https://vo7kov.github.io/js_employees_table_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,113 +1,26 @@
 'use strict';
 
-// Sort table
-const tHead = document.querySelector('thead');
-const tHeadItems = tHead.querySelectorAll('th');
-
-const tBody = document.querySelector('tbody');
-const trsBody = tBody.querySelectorAll('tr');
-
-let lastActiveTrBody;
-let sortedTrsBody;
-let isClickedFirst = false;
-
-tHeadItems.forEach((tHeadItem, index) => {
-  tHeadItem.addEventListener('click', () => {
-    isClickedFirst = !isClickedFirst;
-
-    if (isClickedFirst) {
-      if (index === 0 || index === 1 || index === 2) {
-        sortedTrsBody = [...trsBody].sort((trBody1, trBody2) => {
-          const trBody1String = trBody1.children[index].innerText;
-          const trBody2String = trBody2.children[index].innerText;
-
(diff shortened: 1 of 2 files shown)


TarasHoliuk:
1. Use the image exported from the mobile version in Figma (instead of using huge image and cut it):
image
App: 
image

2. This image shouldn't contain this label:
image

image

TarasHoliuk (changes requested):
You did an amazing job :+1: 
Only a few minor things might be fixed (check comments)

TarasHoliuk:
Would be nice to add `select` input on the `Office` column edit:
image

Implement this part:
image

Do not forget to trim values (when editing and when adding a new row) to avoid adding empty (with spaces only) values

src/scripts/main.js
```diff
@@ -1,3 +1,376 @@
...
+let lastActiveTrBody;
+let sortedTrsBody;
+let isClickedFirst = false;
+
+tHeadItems.forEach((tHeadItem, index) => {
+  tHeadItem.addEventListener('click', () => {
+    isClickedFirst = !isClickedFirst;
+
+    if (isClickedFirst) {
+      if (index === 0 || index === 1 || index === 2) {
```
TarasHoliuk, line 19: do not use magic numbers, save indices to meaningful (with column name) names

src/scripts/main.js
```diff
@@ -1,3 +1,376 @@
...
+
+let lastActiveTrBody;
+let sortedTrsBody;
+let isClickedFirst = false;
+
+tHeadItems.forEach((tHeadItem, index) => {
+  tHeadItem.addEventListener('click', () => {
+    isClickedFirst = !isClickedFirst;
+
+    if (isClickedFirst) {
```
TarasHoliuk, line 18: you're duplicating a big part of your code
try to overwrite it a bit (e.g. handle sort order in every separate case)

src/scripts/main.js
TarasHoliuk, line 19: use `switch`
hint: you can join a few cases, for example:
```
case 0:
case 1:
case 2:
  sortedTrsBody = // rest your code
```

TarasHoliuk (changes requested):

etojeDenys (approved):

PR #436, opened 2023-07-08
Solution

Zibi95:
[DEMO LINK](https://Zibi95.github.io/js_employees_table_DOM/)

### DIFF
--- a/src/scripts/form.js
+++ b/src/scripts/form.js
@@ -1,0 +1,79 @@
+'use strict';
+
+const buildForm = () => {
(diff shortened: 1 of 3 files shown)


DorotaLeniecDev (changes requested):
Please fix the bugs.
- When clicking on the column title, first two clicks are working as expected but then the sorting stops working.
- When dblclicking on the cell the text is acting wierdly, it should remove the text and add input

DorotaLeniecDev (approved):
Good job 💪

PR #437, opened 2023-07-31
Develop

Luk2asz:
- [DEMO LINK](https://Luk2asz.github.io/js_employees_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,199 @@
 'use strict';
 
-// write code here
+const thead = document.querySelector('thead');
+const tbody = document.querySelector('tbody');
+const rows = [...tbody.querySelectorAll('tr')];
+let ascName = true;
+let ascPosition = true;
+let ascOffice = true;
+let ascAge = true;
+let ascSalary = true;
+const body = document.querySelector('body');
+const form = document.createElement('form');
+const notification = document.createElement('div');
+const newRow = document.createElement('tr');
+
+body.appendChild(notification);
+body.appendChild(form);
+
+form.outerHTML = `<form class="new-employee-form">
+<label>Name: <input name="name" type="text" data-qa="name" ></label>
+<label>Position: <input name="position" type="text" data-qa="position" >
+</label>
+<label>Office: <select name="office" data-qa="office">
+<option value="Tokyo">Tokyo</option>
+<option value="Singapore">Singapore</option>
+<option value="London">London</option>
+<option value="New York">New York</option>
+<option value="Edinburgh">Edinburgh</option>
+<option value="San Francisco">San Francisco</option>
+</select>
+</label>
+<label>Age: <input name="age" type="number" data-qa="age" ></label>
+<label>Salary: <input name="salary" type="number"  data-qa="salary" ></label>
+<button name="button">Save to table</button>
+</form>`;
+
+const nameInput = document.querySelector('input[name="name"]');
+const positionInput = document.querySelector('input[name="position"]');
+const officeInput = document.querySelector('select[name="office"]');
+const ageInput = document.querySelector('input[name="age"]');
+const salaryInput = document.querySelector('input[name="salary"]');
+const button = document.querySelector('button');
+
+thead.addEventListener('click', e => {
+  const columnIndex = e.target.cellIndex;
+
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,199 @@
...
+const positionInput = document.querySelector('input[name="position"]');
+const officeInput = document.querySelector('select[name="office"]');
+const ageInput = document.querySelector('input[name="age"]');
+const salaryInput = document.querySelector('input[name="salary"]');
+const button = document.querySelector('button');
+
+thead.addEventListener('click', e => {
+  const columnIndex = e.target.cellIndex;
+
+  if (e.target.cellIndex === 0) {
```
darokrk, line 47: you can refactor this if / else check to be switch function e.g:

```
switch (columnIndex) {
      case 0:
         if (ascName) {
            some code ....
        }
      case 1:
        etc. etc.
}
```

src/scripts/main.js
```diff
@@ -1,3 +1,199 @@
...
+        const current = a.cells[columnIndex].innerHTML;
+        const next = b.cells[columnIndex].innerHTML;
+
+        ascName = false;
+
+        return current.localeCompare(next);
+      });
+    } else {
+      rows.sort((a, b) => {
+        const current = b.cells[columnIndex].innerHTML;
```
darokrk, line 59: maybe we can extract that logic into separate function with passed parameters, like ascName, accPosition, ascOffice etc.? Seems they doing the same oprations.

darokrk (changes requested):
Code works 💪🏽  but we can improve it even more ⚒️ 👨🏽‍💻

src/scripts/main.js
```diff
@@ -1,3 +1,136 @@
...
+});
+
+button.addEventListener('click', (e) => {
+  e.preventDefault();
+  tbody.appendChild(newRow);
+  notification.setAttribute('data-qa', 'notification');
+
+  if (!nameInput.value || !ageInput.value || !officeInput.value
+    || !positionInput.value || !salaryInput.value) {
+    notification.classList.add('error');
```
darokrk, line 99: think you can also extract these logic to the separate function, like `notificationError(text)` you doing the same logic in three places, so will be easier to read to split into smaller ones:

```
function notificationError(text) {
  notification.classList.add('error');
  notification.textContent = text;
}
```

src/scripts/main.js
```diff
@@ -1,3 +1,136 @@
...
+    notification.textContent = 'at least one of the fields is empty';
+  } else if (nameInput.value.length < 4) {
+    notification.classList.add('error');
```
darokrk, line 102: same as above

src/scripts/main.js
```diff
@@ -1,3 +1,136 @@
...
+    notification.textContent = 'Name must have at least 4 letters';
+  } else if (ageInput.value < 18 || ageInput.value > 90) {
+    notification.classList.add('error');
```
darokrk, line 105: same as above

src/scripts/main.js
```diff
@@ -1,3 +1,136 @@
...
+    notification.textContent = 'Age must be between 18 and 90';
+  } else {
+    newRow.outerHTML = `<tr>
+    <td>${nameInput.value}</td>
+    <td>${positionInput.value}</td>
+    <td>${officeInput.value}</td>
+    <td>${ageInput.value}</td>
+    <td>$${Number(salaryInput.value).toLocaleString('en-US')}</td>
+    </tr>`;
+    notification.classList.add('success');
```
darokrk, line 115: logic of that if else block is huge, you can also move that whole operation to separate function like `notificationSuccess()` e.g.

```
function notificationSuccess() {
  notification.textContent = 'You have added a new employee';
    nameInput.value = '';
    positionInput.value = '';
    officeInput.value = 'Tokyo';
    ageInput.value = '';
    salaryInput.value = '';
}
```

darokrk (changes requested):
Good job 💪🏽 code looks now way better, we can even improve it more, look at my last mentions added 🛠️

darokrk (approved):
Good job 👏🏽  looks way better 🔥 👍🏽 !

PR #438, opened 2023-08-04
add task solution

ollavka:
[DEMO LINK](https://ollavka.github.io/js_employees_table_DOM/)


src/scripts/utils/salaryParser.js
```diff
@@ -0,0 +1,18 @@
+'use strict';
+
+const salaryParser = (salary) => {
```
OlEzhA0, line 3: it's not good idea to create parser that can be able to format from number to correct string and from correct string to number. Create function with only 1 task

OlEzhA0 (changes requested):
I can edit your items and save spacings
also I can set age 0 in edit mode (when create, I have an error like age should be between 18 and 90)
Screenshot 2023-08-13 at 7 52 55 PM

src/scripts/modules/form/addNewEmployee.js
```diff
@@ -0,0 +1,56 @@
...
+  const tableRow = document.createElement('tr');
+
+  for (const [key, value] of Object.entries(formData)) {
+    const tableData = document.createElement('td');
+
+    let data = value;
+
+    if (key === 'salary') {
+      data = formatNumberToSalary(+data);
+    }
```
DanilWeda, line 16:
```suggestion
    const salary = key === 'salary';
    const data = salary ? formatNumberToSalary(+data) : value;
```

Alternative variant.

src/scripts/modules/form/addNewEmployee.js
```diff
@@ -0,0 +1,56 @@
...
+  event.preventDefault();
+
+  const form = event.target;
+
+  const tableBody = table.querySelector('tbody');
+  const formData = Object.fromEntries(new FormData(form));
+
+  const errors = getErrors(formData);
+
+  if (errors.length === 0) {
```
DanilWeda, line 37:
```suggestion
  if (!errors.length) {
```

src/scripts/modules/form/createNewEmployeeForm.js
```diff
@@ -0,0 +1,106 @@
...
+  submitButton.textContent = 'Save to table';
+  submitButton.type = 'submit';
+
+  return submitButton;
+};
+
+const createNewEmployeeForm = (domElement) => {
+  const form = document.createElement('form');
+
+  form.className = 'new-employee-form';
```
DanilWeda, line 89: Best way use class list.

src/scripts/modules/notification/getErrors.js
```diff
@@ -0,0 +1,41 @@
...
+  }
+
+  if (wrongAge) {
+    const ageError = {
+      title: 'Wrong age',
+      description: 'The age should be between 18 and 90 years old',
+    };
+
+    errors.push(ageError);
+  }
```
DanilWeda, line 36: Try to think about how to rewrite it. If we have ten errors, we need to use different texts, but one function for push.

src/scripts/modules/notification/pushNotification.js
```diff
@@ -0,0 +1,39 @@
...
+) => {
+  const body = document.body;
+
+  const notifElem = document.createElement('div');
+  const notifTitle = document.createElement('h2');
+  const notifDescription = document.createElement('p');
+
+  notifElem.classList.add('notification', type);
+
+  notifElem.style.top = typeof posTop === 'number'
```
DanilWeda, line 18:
```suggestion
  notifElem.style.top = Number(posTop)
```

src/scripts/modules/table/editTableData.js
```diff
@@ -0,0 +1,113 @@
...
+  return isNumber ? 'number' : 'text';
+};
+
+const saveChanges = (tableData, prevValue, valueToChange) => {
+  const newValue = getNewValue(prevValue, valueToChange);
+
+  if (prevValue.startsWith('$')) {
+    const newSalary = formatNumberToSalary(+newValue);
+
+    if (newSalary.slice(1) === 'NaN') {
```
DanilWeda, line 34: 'NaN' as string? Is it not a mistake? Please check.

src/scripts/modules/table/sortTable.js
```diff
@@ -0,0 +1,88 @@
...
+
+const getParsedData = (firstData, secondData) => {
+  const firstParsedData = parseFloat(firstData.replace(/[^\d.-]/g, ''));
+  const secondParsedData = parseFloat(secondData.replace(/[^\d.-]/g, ''));
+
+  return [firstParsedData, secondParsedData];
+};
+
+const getSortedRows = (rows, sortIndex) => {
+  const sortedRows = [...rows].sort((row1, row2) => {
```
DanilWeda, line 26: Good using spread.

src/scripts/modules/table/sortTable.js
```diff
@@ -0,0 +1,88 @@
...
+
+  const tableTitleIdx = tableTitles.indexOf(sortField);
+
+  if (tableTitleIdx < 0) {
+    return;
+  }
+
+  setSortOrder(tableTitleIdx);
+
+  const tableRows = [...tableBody.rows];
```
DanilWeda, line 61: Spread in this place, and in the inner sort function. You can use spread in one place.

src/scripts/modules/table/sortTable.js
```diff
@@ -0,0 +1,88 @@
...
+
+  const tableTitle = targetItem.textContent;
+
+  const sortedRows = getSortedRowsByField(tableTitle);
+
+  if (!sortedRows) {
+    return;
+  }
+
+  tableBody.append(...sortedRows);
```
DanilWeda, line 85: This spread is required?

src/scripts/utils/salaryParser.js
```diff
@@ -0,0 +1,29 @@
+'use strict';
+
+const formatNumberToSalary = (number) => {
+  if (typeof number !== 'number') {
+    return;
+  }
```
DanilWeda, line 6: Can try to convert, if, for example, you drop '3' in the function?

DanilWeda (changes requested):
Good work! Check some additional things.

PR #439, opened 2023-08-07
Solution

KarolinaWiktoria:
[DEMO](https://KarolinaWiktoria.github.io/js_employees_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,251 @@
 'use strict';
 
-// write code here
+const panel = document.createElement('form');
+const place = document.querySelector('body');
+
+panel.className = 'new-employee-form';
+panel.action = '#';
+panel.method = 'post';
+
+panel.innerHTML = `
+  <label>Name: <input data-qa="name" name="name" type="text"></label>
+  <label>Position: <input
+    data-qa="position" name="position" type="text"></label>
+  <label>Office: <select data-qa="office" name="office">
+    <option>Tokyo</option>
+    <option>Singapore</option>
+    <option>London</option>
+    <option>New York</option>
+    <option>Edinburgh</option>
+    <option>San Francisco</option>
+    </select>
+  </label>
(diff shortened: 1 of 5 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,251 @@
...
+      return value2.localeCompare(value1);
+    }
+
+    return value1.localeCompare(value2);
+  });
+
+  list.append(...result);
+}
+
+addEventListener('click', e => {
```
Radoslaw-Czerniawski, line 140: - i don't see a need for selecting tableFooter
- you could select the tableHeaders by querySelectorAll('th')
image
- `i` is not needed anymore, as the selectedHeader has `cellIndex` property from which we can take appropriate index 
- `k` is also not needed, we may just use the dataset.status for the sorting
- also the sorting either way needs to receive the tableBody so let's just select it there
image

Radoslaw-Czerniawski: Also maybe it would be a good idea to just select the table body at the top of the file as some other functions use it too, let's not repeat such operations unnecessarly

Radoslaw-Czerniawski (changes requested):
Very nice implementation, some refactoring ideas from my side :D

src/scripts/main.js
```diff
@@ -1,3 +1,251 @@
...
+
+  mesage.innerHTML = `
+    <h2 class="title">${title}</h2>
+    <p>${description}</p>
+  `;
+
+  place.append(mesage);
+
+  setTimeout(() => {
+    mesage.hidden = true;
```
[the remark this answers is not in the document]
KarolinaWiktoria: Oh... sure it makes more sense.

src/scripts/main.js
KarolinaWiktoria: Okay thanks! :) i hope now it will be better :D

PR #440, opened 2023-08-19
employees_table_DOM - for testing purposes

RufatRM:
https://RufatRM.github.io/js_employees_table_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,154 @@
 'use strict';
 
-// write code here
+const tableBody = document.querySelector('tbody');
+const rowsArray = Array.from(document.querySelectorAll('tbody tr'));
+const headings = document.querySelectorAll('thead th');
+
+function convertToNumber(string) {
+  return parseInt(string.slice(1).replaceAll(',', ''));
+}
+
+let columnIndex;
+let isAscending = true;
+
+headings.forEach(heading => {
+  heading.addEventListener('click', e => {
+    const index = e.target.cellIndex;
+
+    if (columnIndex === index) {
+      isAscending = !isAscending;
+    } else {
+      columnIndex = index;
+      isAscending = true;
+    }
+
+    rowsArray.sort((a, b) => {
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,154 @@
...
+form.appendChild(nameField);
+form.appendChild(positionField);
+form.appendChild(officeField);
+form.appendChild(ageField);
+form.appendChild(salaryField);
+form.appendChild(submitBtn);
+
+document.body.appendChild(form);
+
+submitBtn.addEventListener('click', () => {
```
Anastasiia-Svintsova, line 135: Avoid using `click` event on the submit button. It's better to attach the event handler to the form's `submit` event. This is because pressing Enter in a text input field also triggers the form's `submit` event. So, change it to form.addEventListener('submit', (e) => { ... })

Anastasiia-Svintsova, line 135: You need to prevent the default form submission using event.preventDefault() inside your function. This is to prevent the page from reloading upon form submission.

src/scripts/main.js
```diff
@@ -1,3 +1,154 @@
...
+
+  const labelText = label.textContent;
+  const textModified = labelText
+    .slice(0, labelText.indexOf(':')).toLowerCase();
+
+  element.dataset.qa = textModified;
+  element.name = textModified;
+  // element.required = true;
+
+  if (`${tag}` === 'select') {
```
Anastasiia-Svintsova, line 104: Avoid using a `for...of` loop to create a new array or append elements to an array. Instead, use `map` or `forEach`.

Anastasiia-Svintsova (changes requested):
Will continue the conversation in the chat

witflash (changes requested):
Good work.

Next steps to improve.

1. Check your notification text and description for any case. For example, if I fill all the fields I have the next message:
image
You check an age but there is no message about it.

2. You should implement editing:
image

PR #443, opened 2023-09-09
js_employees_table_DOM_solution

Lazarend:


SanyaBratashchuk (changes requested):
1. Remove notification when there is no errors 
image
2. Add more specific error messages 
image
3. validation on edit should be the same as on create

Lazarend:
Thanks. I tried to fix it.

etojeDenys (changes requested):
1. you should also validate how user edits the fields
image

Lazarend:
Added validation to edited cells.

etojeDenys (approved):
good job

PR #445, opened 2023-09-15
solve the task

eugene-kulish:
[DEMO LINK](https://eugene-kulish.github.io/js_employees_table_DOM/)


sTorba24 (changes requested):
Well done!
Lets improve it

make error messages more meaningful.
Short? how should be
Age not valid? Which valid?
etc
image

image

sTorba24 (approved):
Much better!)

PR #446, opened 2023-09-26
first

Vladimir-Zadorozhnyi:
https://Vladimir-Zadorozhnyi.github.io/js_employees_table_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,176 @@
 'use strict';
 
-// write code here
+const table = document.querySelector('table');
+const headers = table.querySelectorAll('thead th');
+const list = table.querySelector('tbody');
+let rows = table.querySelectorAll('tbody tr');
+const orders = {};
+
+headers.forEach((element, index) => {
+  orders[index] = 'asc';
+
+  element.addEventListener('click', () => {
+    const current = `td:nth-child(${index + 1})`;
+    const order = orders[index];
+
+    if (order === 'asc') {
+      orders[index] = 'desc';
+    } else {
+      orders[index] = 'asc';
+    }
+
+    function replace(num) {
+      return num.querySelector(current)
+        .textContent.replace(/[$,]/g, '');
+    }
+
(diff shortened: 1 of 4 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,176 @@
...
+  } else if (positionEmployee.length === 0) {
+    showNotification('error', 'The "Position" is empty.');
+  } else if (!officeEmployee) {
+    showNotification('error', 'The Office field is not selected.');
+  } else if (isNaN(ageEmployee) || ageEmployee < 18 || ageEmployee > 90) {
+    showNotification('error',
+      'The age must be a valid number between 18 and 90.');
+  } else if (isNaN(salaryEmployee) || salaryEmployee < 0) {
+    showNotification('error', 'The "Salary" is empty.');
+  } else {
```
anastasiia-tilikina, line 132: too many else if! refactor it please.
you can write like this

```javascript
if (cond) {
  // do something
  return
}

//etc.
```

anastasiia-tilikina (changes requested):
Check my comment.
Also, your row selection doesn't work. please fix it

src/scripts/main.js
```diff
@@ -1,3 +1,185 @@
...
+
+  if (isNaN(ageEmployee) || ageEmployee < 18 || ageEmployee > 90) {
+    return showNotification('error',
+      'The age must be a valid number between 18 and 90.');
+  }
+
+  if (isNaN(salaryEmployee) || salaryEmployee < 0) {
+    return showNotification('error', 'The "Salary" is empty.');
+  } else {
+    const newRow = list.insertRow();
```
Viktor-Kost, line 142: no needs to use `else` if you use early `return`
```suggestion
  if (isNaN(salaryEmployee) || salaryEmployee < 0) {
    return showNotification('error', 'The "Salary" is empty.');
  } 
  
  const newRow = list.insertRow();
```

Viktor-Kost (approved):

PR #447, opened 2023-09-28
first  solution

spencerspe:
- [DEMO LINK](https://spencerspe.github.io/js_employees_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,175 @@
 'use strict';
 
 // write code here
+const table = document.querySelector('table');
+const tableBody = document.querySelector('tbody');
+const tablerows = document.querySelectorAll('tbody tr');
+const tablerowsArray = Array.from(tablerows);
+
+const offices = [`Tokyo`, `Singapore`,
+  `London`, `New York`, `Edinburgh`, `San Francisco`];
+
+const formFields = {
+  'name': 'text',
+  'position': 'text',
+  'office': 'select',
+  'age': 'number',
+  'salary': 'number',
+};
+
+function capitalise(string) {
+  return string.charAt(0).toUpperCase() + string.slice(1);
+}
+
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,175 @@
...
+
+  if (evt.target.tagName === 'TD') {
+    evt.stopPropagation();
+    tablerows.forEach(rowClass => rowClass.classList.remove('active'));
+    evt.target.parentNode.classList = 'active';
+    evt.target.contentEditable = true;
+  }
+});
+
+// form
```
witflash, line 60: You can avoid using such comments because the next constant exactly is `form` :)

src/scripts/main.js
```diff
@@ -1,3 +1,175 @@
...
+
+  if (enteredAge < 18 || enteredAge > 90) {
+    message = 'Age entry must be between 18 years and 90 years';
+    messageClass = 'error';
+  }
+
+  notificationTitle.textContent = message;
+  notification.classList.add(messageClass);
+  notification.append(notificationTitle);
+  form.append(notification);
```
witflash, line 152: Don't append a notification inside your form, because it will look weird:
image

You can use `form.after(notification)` to render your notification as a form sibling but not a child.

witflash (changes requested):
Good job!
Pay attention to my comments, please

lerastarynets (changes requested):
For now user is able to add a row with name consisting of whitespaces only and salary <0
pls fix)
Screenshot at Oct 04 14-44-44

anastasiia-tilikina (changes requested):
Ooops!)

Seems like something went wrong with the logic)

image

Pls fix it

witflash (approved):
Great!

I see you've corrected all the problems!
I like your work. Well done 👍

PR #450, opened 2023-10-14
Sort Table is ready

vanya-kalyenichenko:


lerastarynets (changes requested):
I see you did a hard work) Let's improve it a little bit

error message is unclear, it's impossible to understand what exactly is wrong, pls provide specific error messages for all the field
Screenshot at Oct 14 13-17-26

when editing form, one can enter invalid values, pls fix
Screenshot at Oct 14 13-18-10

vanya-kalyenichenko:
About age and name I can add special require, but for position I don't know, it`s impossible, because we don't have list of jobs... and the name of job could be so different

volodymyr-soltys97 (approved):
🔥

PR #453, opened 2023-10-26
Solution

andrii-rykal:
[DEMO LINK](https://andrii-rykal.github.io/js_employees_table_DOM/)


src/scripts/main.js
```diff
@@ -1,3 +1,310 @@
 'use strict';
 
-// write code here
+const NAME = 'Name';
+const POSITION = 'Position';
+const OFFICE = 'Office';
+const AGE = 'Age';
+const SALARY = 'Salary';
```
VitaliyBondarenko1982, line 7: suggest to make one constant as object and each constant will be as field of object
```
const FORM_FIELD = {
  NAME: 'name',
  ....
}

File: src/scripts/main.js [hunk 2]
```diff
@@ -1,3 +1,310 @@
...
+const messageName = 'Поле "Name:" повинно мати не менше 4 символів';
+const messagePosition = 'Необхідно вказати посаду';
+const messageFullData = 'Недостатньо даних';
+const messageAge = 'Вік від 18 до 90';
+const messageSuccess = 'Співробітника успішно додано';
```
VitaliyBondarenko1982, line 12: all possible notification messages also better to wrtite in object.

src/scripts/main.js
```diff
@@ -1,3 +1,310 @@
...
+
+function informMessage(type, text) {
+  const div = document.createElement('div');
+
+  div.className = 'notification';
+  div.classList.add(type);
+  div.dataset.qa = 'notification';
+  div.innerText = text;
+  div.style.color = 'black';
+  form.prepend(div);
```
VitaliyBondarenko1982, line 305:
```suggestion
  document.body.append(div);
```
it can make better view for notification
now it is not good
chrome-capture-2023-9-26

VitaliyBondarenko1982 (changes requested):
Good job.
Suggest some improvements.
- after click on some sort field I can't to edit any cell
- try to prevent some jumping row when start ediging
chrome-capture-2023-9-26 (1)

anastasiia-tilikina (approved):
GJ!

PR #468, opened 2023-12-13
done

victorgri:
https://victorgri.github.io/js_employees_table_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,260 @@
+/* eslint-disable max-len */
 'use strict';
 
-// write code here
+const QUERY = {
+  'Name': 0,
+  'Position': 1,
+  'Office': 2,
+  'Age': 3,
+  'Salary': 4,
+};
+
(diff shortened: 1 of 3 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,260 @@
...
+
+let desc = false;
+let query = '';
+
+function sortEmployees(sortedQuery) {
+  switch (sortedQuery) {
+    case queryName:
+    case queryPosition:
+    case queryOffice:
+      return desc
```
VitaliyBodnar, line 34: Let's improve your solution
- don't need to write logic for asc and desc in every case. 
- you can create variable, set sorters list to this variable 
- if sort type is desc you need to rerun a reversed sorted list (with method revers)
- if sort type is asc than just return a sorted list

VitaliyBodnar (changes requested):

VitaliyBodnar (changes requested):
.

PR #470, opened 2023-12-25
add task solution

Viktoriia6666:
- [DEMO LINK](https://Viktoriia6666.github.io/js_employees_table_DOM/)


DarkMistyRoom (changes requested):
Good, but a few comments to editing:

- On blur save changes to table cell. Remove input and set new text.
- Make the input focused after double-clicking
- Show validation messages on editing
- If an input is empty on submitting return initial value.

PR #478, opened 2024-01-28
solution

vinogradova8:
[DEMO LINK](https://vinogradova8.github.io/js_employees_table_DOM/)


src/scripts/main.js
```diff
@@ -1,3 +1,358 @@
...
+  const notification = document.createElement('div');
+
+  setTimeout(() => {
+    notification.remove();
+  }, 2000);
+
+  notification.dataset.qa = 'notification';
+  notification.classList.add('notification');
+
+  if (employeeName.length < 4 || age < 18 || age > 90) {
```
lerastarynets, line 187: consider moving numbers to consts with consistent naming and rreuse them where needed

lerastarynets (changes requested):
I'm redirected to page with error 405 when i click on save to table button, also when editing, I can add invalid values to table
Screenshot at Jan 29 10-30-01

Viktor-Kost (changes requested):
We should show to our user what exactly was wrong
So
1. do not reset form if data is invalid
2. show different notifications for different error cases

https://github.com/mate-academy/js_employees_table_DOM/assets/117587463/6f7f6fc5-b659-45d8-a0ef-31ba260fe822

Screenshot 2024-01-29 at 23 23 42

etojeDenys (approved):
good job

PR #480, opened 2024-02-01
add task solution

alina-kabanets:
https://alina-kabanets.github.io/js_employees_table_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,230 @@
 'use strict';
 
-// write code here
+const body = document.querySelector('body');
+const table = document.querySelector('table');
+const tbody = document.querySelector('tbody');
+const header = document.querySelector('thead');
+let rows = [...tbody.querySelectorAll('tr')];
+const theads = [...header.querySelectorAll('th')];
+
+header.addEventListener('click', (e) => {
+  const target = e.target.closest('th');
+  const index = e.target.cellIndex;
+
+  const sorting = function() {
+    rows.sort((a, b) => {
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,230 @@
...
+      case 'salary':
+        input.type = 'number';
+        break;
+    }
+  } else {
+    const select = document.createElement('select');
+
+    const locations = [
+      'Tokyo', 'Singapore', 'London', 'New York', 'Edinburgh', 'San Francisco',
+    ];
```
Viktor-Kost, line 113: move this array outside the loop

src/scripts/main.js
```diff
@@ -1,3 +1,230 @@
...
+  const age = formData.get('age');
+  const salary = formData.get('salary');
+
+  if (newName.length < 4) {
+    showNotification('warning', 'Name can not be less than 4 characters');
+
+    return;
+  };
+
+  if (age < 18 || age > 90) {
```
Viktor-Kost, line 152: move numeric values to variables to avoid using `magic` numbers

src/scripts/main.js
```diff
@@ -1,3 +1,230 @@
...
+      return true;
+    }
+  };
+
+  function handleInput() {
+    if (isEmpty()) {
+      cellInput.value = initialValue;
+    } else {
+      changedCell.innerText = cellInput.value;
+    }
```
Viktor-Kost, line 218: you can use early return here instead of `if/else`

Viktor-Kost (approved):
Great job!

PR #487, opened 2024-03-02
mySolution

AntonFof:
[DEMO LINK](https://AntonFof.github.io/js_employees_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,194 @@
 'use strict';
 
-// write code here
+const tbody = document.querySelector('tbody');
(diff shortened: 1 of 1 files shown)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
The name and position cannot be an empty, need to fix it
image

AntonFof:
I cannot add a new row if "name" and "position" are empty.  How could you do it?

SanyaBratashchuk (changes requested):
Looks like editing is not implemented

PR #498, opened 2024-05-09
Done

DenisGurskiy:
https://denisgurskiy.github.io/js_employees_table_DOM/


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code.
The name and position cannot be empty, you need to fix it 
image

volodymyr-soltys97 (changes requested):
Almost done!
You need to fix the error text 
image
image
image

anastasiiavorobiova (approved):
Great work!

PR #505, opened 2024-05-29
Added task solution

OleksiiNesteruk:
- [DEMO LINK](https://OleksiiNesteruk.github.io/js_employees_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,203 @@
 'use strict';
 
-// write code here
+const table = document.querySelector('table');
+
+const tableHeaders = table.querySelectorAll('th');
+const tbody = table.tBodies[0];
+const rows = Array.from(tbody.rows);
+const sortDirection = Array.from(tableHeaders).map(() => true);
+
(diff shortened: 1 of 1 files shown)


anastasiia-tilikina (changes requested):
GJ

We need to fix some issues here:
- I'm not able to select added row
- please implement correctly sortin requirements

> When users clicks on one of the table headers, table should be sorted in ASC order, the second click sorts it in DESC order.
When users click on a new title, always sort in ASC order.

OleksiiNesteruk:
The demo link is not working. Could you please help us fix it?

volodymyr-soltys97 (changes requested):
Hi, the comments from the previous review still not fixed, also you need to make a deploy 
image

lerastarynets (changes requested):
Pls validate salary field, so that I can't submit values less than 0
Screenshot at May 30 15-33-07

lerastarynets (approved):
Keep up the great work!

PR #508, opened 2024-06-10
Solution

mukutiuk:
https://mukutiuk.github.io/js_employees_table_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,303 @@
 'use strict';
+'use strict';
 
-// write code here
+const tbody = document.querySelector('tbody');
+const title = document.querySelector('thead');
+
+const rows = [...tbody.querySelectorAll('tr')];
+let revers = '';
+
+title.addEventListener('click', (e) => {
+  const value = e.target.closest('th').textContent.toLowerCase();
+
+  const intex = e.target.cellIndex;
+  const cells = [];
+
+  tbody.querySelectorAll('tr').forEach((item) => {
+    cells.push(item.cells[intex].textContent);
+  });
+
+  const newRows = rows.sort((a, b) => {
+    const cellsA = a.cells[intex].textContent;
+    const cellsB = b.cells[intex].textContent;
+
+    switch (value) {
+      case 'name':
+      case 'position':
+        return cellsA.localeCompare(cellsB);
+
+      case 'age':
+        return cellsA - cellsB;
+
+      case 'salary':
+        return (
+          +cellsA.replace(',', '').slice(1) - +cellsB.replace(',', '').slice(1)
+        );
+
+      default:
+        return null;
+    }
+  });
+
+  if (value === revers) {
+    newRows.reverse();
+    revers = '';
+  } else {
+    revers = value;
+  }
+
+  tbody.innerHTML = '';
+
+  newRows.forEach((item) => {
+    tbody.appendChild(item);
+  });
+});
+
+const handleActiveClas = () => {
+  tbody.querySelectorAll('tr').forEach((item) => {
+    item.addEventListener('click', (e) => {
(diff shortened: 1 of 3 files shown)


volodymyr-soltys97 (changes requested):
Hi, the new employee is not added to the table
image
image

anastasiiavorobiova (changes requested):
Great work! 
Screenshot 2024-06-11 at 13 41 24
Can't add an employee to the table. Consider fixing it

GUSILLUS (changes requested):
Hi! 

1. Let's fix the add menu placement it needs to be on one level with the table
2. Let's add some logic to check if we want to edit age or salary input and change it edit input type to number to avoid using a string in this field
3. Let's fix the width of the edit input because even in small cells we have a big edit input
4. Catch some bug, if we add some employer and try to sort the table he will remove
image

src/scripts/main.js
```diff
@@ -1,3 +1,323 @@
...
+});
+
+const td = document.querySelectorAll('td');
+
+td.forEach((item) => {
+  item.addEventListener('dblclick', (e) => {
+    const findInput = document.querySelector('.cell-input');
+    const oldValue = e.target.textContent;
+
+    console.log();
```
anastasiia-tilikina, line 226: remove console. Lint tests are failed

anastasiia-tilikina (changes requested):
I still can add not valid values when I edit fields.
let's add some validation

src/scripts/main.js
```diff
@@ -1,3 +1,364 @@
 'use strict';
+'use strict';
+
+const tbody = document.querySelector('tbody');
+const title = document.querySelector('thead');
+let revers = '';
+
+const handleActiveClas = () => {
```
DarkMistyRoom, line 8:
```suggestion
const handleActiveClass = () => {
```

src/scripts/main.js
```diff
@@ -1,3 +1,364 @@
...
+  div.setAttribute('class', 'notification');
+
+  div.style.position = 'absolute';
+
+  if (
+    inputAge.value < 18 ||
+    inputAge.value > 90 ||
+    inputName.value.length < 4 ||
+    inputPosition.value.trim().length === 0
+  ) {
```
DarkMistyRoom, line 147: would be great to create a validation function and move a part of this logic there

DarkMistyRoom: try to break your code into functions

DarkMistyRoom (changes requested):
Very good! Let's improve a bit:
- the Office sorting works only after a second click
- I can add an employee with only white spaces in the name. Please fix it
image

- I can't select or edit an added employee
- Put the current value in the input on editing
- show the same error messages on editing
- use the same select for editing Office as in the form

lerastarynets (changes requested):
Editing is still unavailable, also i can add invalid salary values
Pls ask for help in **fe_chat** if you don't understand something
Screenshot at Jun 12 13-39-03

vadiimvooo (changes requested):
Good job, but there is one bug. I can not edit office field, please watch [video](https://www.loom.com/share/e43f4d65ddab4178ac9fc2c0e25c0c15). Feel free to ask for help.

DarkMistyRoom (changes requested):
- place current value into the input on editing
- validate values on editing the same as on adding a new employee (it's possible to set name with less then 4 letters)

DarkMistyRoom (changes requested):
Everything is great! Only one thing - I can't edit a salary

SanyaBratashchuk (changes requested):
Add validation on salary editing  as well
image

SanyaBratashchuk (approved):
LGTM

PR #520, opened 2024-06-20
task solution

AlinaOvod:
[DEMO LINK](https://alinaovod.github.io/js_employees_table_DOM/)


anastasiia-tilikina (changes requested):
Good job!

One of your tests is failed now
image

this is because native HTML validation blocks your custom validation 
I would suggest you to remove required attributes since you're using your own validation. This should fix your tests

Also cells editing is not working now

volodymyr-soltys97 (changes requested):
Hi, you need to make a deploy and check it after on the demo link.
Task description 
image

image

anastasiiavorobiova (approved):
Great work!

PR #523, opened 2024-06-21
js_employees_table_DOM

VazilX:
[DEMO LINK](https://VazilX.github.io/js_employees_table_DOM/)


SanyaBratashchuk (changes requested):
it should allow to edit only one row at a time 
image

SanyaBratashchuk (changes requested):
1. should be select on edit as on create 
image
2. notifications shouldn't overlap each other 
image

image

TarasHoliuk:
It would be nice to format salary when editing as well

image

But everything else looks good - approved ✔️

TarasHoliuk (approved):

PR #532, opened 2024-06-26
my solution

viktalfrontend:


volodymyr-soltys97 (changes requested):
Good job! 
Let's make your code better
The name and position cannot be only spaces, fix it
image

volodymyr-soltys97 (approved):
Well done! 🔥

PR #534, opened 2024-06-26
My solution to the task Employees table

Karanelus:
[DEMO LINK](https://karanelus.github.io/js_employees_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,261 @@
 'use strict';
 
-// write code here
+const pageBody = document.querySelector('body');
+const tableHeader = document.querySelector('thead');
+const tableHeaderCells = tableHeader.querySelector('tr');
+const tableBody = document.querySelector('tbody');
+let isSecondClick = false;
+const toNumber = (string) => {
+  if (typeof string === 'string' && string.length !== 0) {
+    const clearNum = string.replace(/[$, ]/g, '');
+
+    return parseInt(clearNum, 10);
+  }
+};
+let clickedSortTable;
+
+tableHeaderCells.addEventListener('click', (e) => {
+  const headerChildren = [...tableHeaderCells.children];
+  const columnName = e.target.closest('th');
+
+  if (columnName !== null) {
+    const sortedColumn = headerChildren.indexOf(columnName);
+    const columnsToSort = [...tableBody.children];
+
+    if (clickedSortTable === columnName) {
+      isSecondClick = !isSecondClick;
+    } else {
+      isSecondClick = false;
+    }
+
+    clickedSortTable = columnName;
+
+    const columnsSorted = columnsToSort.sort((a, b) => {
+      const elemA = a.children[sortedColumn].innerHTML;
+      const elemB = b.children[sortedColumn].innerHTML;
+
+      const aToNum = toNumber(elemA);
+      const bToNum = toNumber(elemB);
+
+      if (Number.isNaN(aToNum)) {
+        return !isSecondClick
+          ? elemA.localeCompare(elemB)
+          : elemB.localeCompare(elemA);
+      }
+
+      return !isSecondClick ? aToNum - bToNum : bToNum - aToNum;
+    });
+
+    tableBody.innerHTML = '';
+
+    columnsSorted.forEach((el) => {
+      tableBody.appendChild(el);
+    });
+  }
+});
+
+tableBody.addEventListener('click', (e) => {
+  const headerChildren = [...tableBody.children];
+  const columnName = e.target.closest('tr');
+
+  headerChildren.forEach((el) => el.removeAttribute('class'));
+  columnName.classList.add('active');
+});
+
+const formContainer = document.createElement('form');
+const nameLabel = document.createElement('label');
+const nameInput = document.createElement('input');
+const posiitionLabel = document.createElement('label');
+const posiitionInput = document.createElement('input');
+const officeLabel = document.createElement('label');
+const officeChoose = document.createElement('select');
+const ageLabel = document.createElement('label');
+const ageInput = document.createElement('input');
+const salaryLabel = document.createElement('label');
+const salaryInput = document.createElement('input');
+const formButton = document.createElement('button');
+const cityOptions = [
+  `Tokyo`,
+  `Singapore`,
+  `London`,
+  `New York`,
+  `Edinburgh`,
+  `San Francisco`,
+];
+const inputs = [nameInput, posiitionInput, officeChoose, ageInput, salaryInput];
+
+const modifyElement = (parentEl, child, cellName, type) => {
+  parentEl.innerHTML = `${cellName}:`;
+  parentEl.setAttribute('data-qa', cellName.toLowerCase());
+  child.required = true;
+  child.type = type;
+  parentEl.appendChild(child);
+};
+
+modifyElement(nameLabel, nameInput, 'Name', 'text');
+modifyElement(posiitionLabel, posiitionInput, 'Position', 'text');
+modifyElement(ageLabel, ageInput, 'Age', 'number');
+modifyElement(salaryLabel, salaryInput, 'Salary', 'number');
+
+formContainer.classList.add('new-employee-form');
+
+officeLabel.innerHTML = 'Office:';
+officeLabel.setAttribute('data-qa', 'office');
+officeLabel.appendChild(officeChoose);
+officeChoose.required = true;
+formButton.innerText = 'Save to table';
+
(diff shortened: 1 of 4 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,263 @@
...
+
+    columnsSorted.forEach((el) => {
+      tableBody.appendChild(el);
+    });
+  }
+});
+
+tableBody.addEventListener('click', (e) => {
+  const rowsChildren = [...tableBody.children];
+  const columnName = e.target.closest('tr');
```
Zibi95, line 61: Rename `columnName` to `selectedRow` for better readability.

src/scripts/main.js
```diff
@@ -1,3 +1,263 @@
...
+
+  rowsChildren.forEach((row) => row.removeAttribute('class'));
```
Zibi95, line 63: Iterating over all elements to remove the class is not the most efficient approach. We can improve this by keeping track of the currently active row and only removing the class from that specific row. And we can use for that purpose method `classList.remove()`

src/scripts/main.js
```diff
@@ -1,3 +1,263 @@
...
+const nameLabel = document.createElement('label');
+const nameInput = document.createElement('input');
+const positionLabel = document.createElement('label');
+const positionInput = document.createElement('input');
+const officeLabel = document.createElement('label');
+const officeChoose = document.createElement('select');
+const ageLabel = document.createElement('label');
+const ageInput = document.createElement('input');
+const salaryLabel = document.createElement('label');
+const salaryInput = document.createElement('input');
```
Zibi95, line 77: Create a helper function to encapsulate the repetitive element creation logic. For example, you can create a function called `createFormField` and keep there logic regarding creation and applying some attributes.

src/scripts/main.js
```diff
@@ -1,3 +1,263 @@
...
+});
+
+formContainer.appendChild(nameLabel);
+formContainer.appendChild(positionLabel);
+formContainer.appendChild(officeLabel);
+formContainer.appendChild(ageLabel);
+formContainer.appendChild(salaryLabel);
+formContainer.appendChild(formButton);
+
+pageBody.appendChild(formContainer);
```
Zibi95, line 126: Always try to keep logic in functions and name it properly so its easy to follow what that function should do. Thanks to that our script is easier to follow.

src/scripts/main.js
```diff
@@ -1,3 +1,263 @@
...
+  containerTitle.innerText = title;
+  containerDescr.innerText = description;
+
+  container.appendChild(containerTitle);
+  container.appendChild(containerDescr);
+  pageBody.appendChild(container);
+
+  setTimeout(() => {
+    container.style.display = 'none';
+  }, 3000);
```
Zibi95, line 149: Consider making the timeout duration configurable.

src/scripts/main.js
```diff
@@ -1,3 +1,263 @@
...
+  const containerTitle = document.createElement('h2');
+  const containerDescr = document.createElement('p');
+
+  container.setAttribute('data-qa', 'notification');
+  containerTitle.classList.add('title');
+
+  if (type === 'success' || type === 'error') {
+    container.classList.add('notification', type);
+  }
+  containerTitle.classList.add('title');
```
Zibi95, line 139: This is doubled. In line `134` you have done that already.

src/scripts/main.js
```diff
@@ -1,3 +1,263 @@
...
+  containerTitle.innerText = title;
```
Zibi95, line 140: Using `innerText` has some downsides, such as performance issues and different behaviors across browsers. Using `textContent` is generally preferred for setting text.

src/scripts/main.js
```diff
@@ -1,3 +1,263 @@
...
+    ageInput.value === '' ||
+    ageInput.value < MIN_AGE ||
+    ageInput.value > MAX_AGE ||
+    salaryInput.value === ''
+  ) {
+    return pushNotification(
+      'Error',
+      'Please, fix all the problems in form',
+      'error',
+    );
```
Zibi95, line 167: Make notification messages more descriptive.

src/scripts/main.js
```diff
@@ -1,3 +1,263 @@
...
+formButton.addEventListener('click', (e) => {
+  e.preventDefault();
+
+  if (
+    nameInput.value === '' ||
+    positionInput.value === '' ||
+    ageInput.value === '' ||
+    ageInput.value < MIN_AGE ||
+    ageInput.value > MAX_AGE ||
+    salaryInput.value === ''
```
Zibi95, line 161: There is missing logic in checking if the inputs are valid. 

> If Name value has less than 4 letters, show error notification.

src/scripts/main.js
```diff
@@ -1,3 +1,263 @@
...
+
+    edCellInput.type = indexOfCell < 3 ? 'text' : 'number';
+    edCellInput.value = cellInitialtName;
+    edCellLabel.appendChild(edCellInput);
+    edCellForm.setAttribute('previous', cellInitialtName);
+    edCellForm.setAttribute('cellType', cellInfo[indexOfCell]);
+    edCellForm.appendChild(edCellLabel);
+
+    editedCell.textContent = '';
+    editedCell.appendChild(edCellForm);
```
Zibi95, line 223: There is so many repetitive code for creating elements. Consider adding helper function for this.

src/scripts/main.js
```diff
@@ -1,3 +1,263 @@
...
+    const parentElem = formHandle.parentElement;
+    const formAttributes = [...formHandle.attributes];
+    const formPreviousName = formAttributes.find(
+      (el) => el.name === 'previous',
+    );
+    const formCellType = formAttributes.find((el) => el.name === 'celltype');
+    const formCellTypeSalary = formCellType.value === 'Salary';
+    const inputElem = formHandle.querySelector('input');
+
+    formHandle.addEventListener('submit', (ev) => {
```
Zibi95, line 242: Ensure the form submission is correctly handled without attaching a new event listener each time. We can attach one listener on the form outside of the function.

src/scripts/main.js
```diff
@@ -1,3 +1,263 @@
...
+      ev.preventDefault();
+
+      if (inputElem.value.trim() !== '') {
+        parentElem.textContent = '';
+
+        if (formCellTypeSalary) {
+          const inputNumber = parseInt(inputElem.value);
+
+          parentElem.innerText = `$${inputNumber.toLocaleString('en-US')}`;
```
Zibi95, line 251: Ensure to set with `textContent` for better performance and consistency.

Zibi95 (changes requested):
There is a lot of repetitive code. Follow the suggestions to make this code better! By following these comments, you will make the code cleaner, more efficient, and easier to maintain. Great job on implementing the table functionality! Keep up the good work!

src/scripts/main.js
```diff
@@ -1,3 +1,298 @@
 'use strict';
 
-// write code here
+const MIN_AGE = 18;
+const MAX_AGE = 100;
```
Zibi95, line 4: One test is failing as the MAX_AGE should be 90
image

src/scripts/main.js
```diff
@@ -1,3 +1,298 @@
...
+
+  return label;
+};
+
+const getElInput = (element) => {
+  return element.firstElementChild;
+};
+
+const formContainer = document.createElement('form');
+const nameElement = createElement('Name', 'text');
```
Zibi95, line 61: Use here even more descriptive name like `nameFormField`

src/scripts/main.js
```diff
@@ -1,3 +1,298 @@
...
+const formButton = document.createElement('button');
+const inputs = [
+  getElInput(nameElement),
+  getElInput(positionElement),
+  officeChoose,
+  getElInput(ageElement),
+  getElInput(salaryElement),
+];
+
+formContainer.classList.add('new-employee-form');
```
Zibi95, line 76: The logic for creating the form should be within the function body, as this will make the code more meaningful. If the function is properly named and focuses on a single task, reading the code will be like reading a book.

Zibi95 (changes requested):
Well done 👍. Fix one test case and checkout my comments left below.

src/scripts/main.js
Karanelus: Where can i read test requirements for the future tasks?

natalia-klonowska: usually it's mention in task instruction 
image
you can also check tests in cypress/integration

natalia-klonowska (approved):

PR #537, opened 2024-06-28
solution

VSkits:
[DEMO LINK](https://VSkits.github.io/js_employees_table_DOM/)


anastasiiavorobiova (changes requested):
Great work! Keep doing! Consider fixing your tests. The sort direction should be opposite. Also, the user should be able to see notifications if the form isn't filled properly.

PR #539, opened 2024-06-29
solution

VSkits:
[DEMO LINK](https://VSkits.github.io/js_employees_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,374 @@
 'use strict';
 
-// write code here
+const table = document.querySelector('table');
+const tbody = table.querySelector('tbody');
+
+// #region sorting
+
+const headings = document.querySelectorAll('th');
+const footerHeadings = document.querySelectorAll('tfoot th');
+const rows = [...tbody.querySelectorAll('tr')];
+
+const sortDirections = Array.from({ length: headings.length }, () => 1);
+
+headings.forEach((heading, index) => {
+  heading.addEventListener('click', () => {
+    sortColumn(index);
+  });
+});
+
+footerHeadings.forEach((footerHeading, index) => {
+  footerHeading.addEventListener('click', () => {
+    sortColumn(index);
+  });
+});
+
+function sortColumn(index) {
+  sortDirections[index] *= -1;
+
+  rows.sort((a, b) => {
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,374 @@
...
+const table = document.querySelector('table');
+const tbody = table.querySelector('tbody');
+
+// #region sorting
+
+const headings = document.querySelectorAll('th');
+const footerHeadings = document.querySelectorAll('tfoot th');
+const rows = [...tbody.querySelectorAll('tr')];
+
+const sortDirections = Array.from({ length: headings.length }, () => 1);
```
vadiimvooo, line 12: Just fix it to pass all tests
```suggestion
const sortDirections = Array.from({ length: headings.length }, () => -1);
```

src/scripts/main.js
```diff
@@ -1,3 +1,374 @@
...
+  const age = +inputToCheck['age'].value;
+  const position = inputToCheck['position'].value;
+  const office = inputToCheck['office'].value;
+  const salary = inputToCheck['salary'].value;
+
+  if (position === '' || office === '' || salary === '') {
+    return { type: 'error', text: 'All fields must be filled in' };
+  }
+
+  if (nameValue.length < 4) {
```
vadiimvooo, line 247: Use constant to save such variables

vadiimvooo (changes requested):
Good job, but you have to pass all tests.

src/scripts/main.js
VSkits: let`s go!!! thanks)))

volodymyr-soltys97 (changes requested):
Good job!
To improve:
1. The name and position cannot be only spaces, need to fix it
image

anastasiiavorobiova (changes requested):
Great work! The new element disappears after sorting the table

VSkits:
> Great work! The new element disappears after sorting the table

how i can fix this? preventDefault sould add to editing table block?

anastasiiavorobiova (changes requested):
Make sure to redeploy your solution, because the demo is still working the same

anastasiiavorobiova (approved):
Awesome work!

PR #540, opened 2024-06-30
add test

olehmarushchak:
https://olehmarushchak.github.io/js_employees_table_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,5 +1,3 @@
-/* eslint-disable no-shadow */
-/* eslint-disable prettier/prettier */
 /* eslint-disable max-len */
 'use strict';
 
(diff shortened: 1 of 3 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,384 @@
...
+    newCell.textContent = newMember[key];
+    newRow.appendChild(newCell);
+  }
+
+  staff.push(newMember);
+  tbody.appendChild(newRow);
+
+  table = tbody.querySelectorAll('tr');
+
+  console.log(table, staff);
```
volodymyr-soltys97, line 379: Remove console.log
```suggestion
```

src/scripts/main.js
```diff
@@ -1,3 +1,384 @@
+/* eslint-disable no-shadow */
+/* eslint-disable prettier/prettier */
+/* eslint-disable max-len */
```
volodymyr-soltys97, line 3: Don't need to disable eslint rules
```suggestion
```

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
The name and position cannot be only spaces, you need to fix it 
image

PR #541, opened 2024-06-30
solution

hsvirina:
https://hsvirina.github.io/js_employees_table_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,161 @@
 'use strict';
 
-// write code here
+const body = document.querySelector('body');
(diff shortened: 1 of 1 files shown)


volodymyr-soltys97 (changes requested):
GJ!
Let's improve your code
image

lerastarynets (changes requested):
Almost there, just one problem left)
Salary can't be less than 0
Screenshot at Jul 01 11-49-46

lerastarynets (approved):
You're making excellent progress!

PR #550, opened 2024-07-07
add solution of task js_employees_table_DOM

Voloshynd:
- [DEMO LINK](https://Voloshynd.github.io/js_employees_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,407 @@
 'use strict';
 
-// write code here
+const table = document.querySelector('table');
+
+let sortedRows = [];
+
+table.tHead.addEventListener('click', (e) => {
+  sortEmployees(e.target.textContent, e.target);
+});
+
+selectRow();
+
+function getIdx(title) {
+  const idx = 0;
+
+  switch (title) {
+    case 'Position':
+      return 1;
+    case 'Office':
+      return 2;
+    case 'Age':
+      return 3;
+    case 'Salary':
+      return 4;
+    default:
+      return idx;
+  }
+}
+
+function sortEmployees(title, elem) {
+  const idx = getIdx(title);
+
+  switch (title) {
+    case 'Age':
+    case 'Salary':
+      if (elem.className === '' || elem.className === 'DESC') {
+        sortedRows = [...table.tBodies[0].children].sort((a, b) => {
+          if (
+            +[...a.children][idx].textContent.replace(/[^0-9.]/g, '') >
+            +[...b.children][idx].textContent.replace(/[^0-9.]/g, '')
+          ) {
+            return 1;
+          }
+
+          if (
+            +[...a.children][idx].textContent.replace(/[^0-9.]/g, '') <
+            +[...b.children][idx].textContent.replace(/[^0-9.]/g, '')
+          ) {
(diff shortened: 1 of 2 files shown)


volodymyr-soltys97 (changes requested):
Hi, you need to add data attributes for tests, also if you have problem with tests feel free in the fe_chat 
image

src/scripts/main.js
```diff
@@ -1,3 +1,422 @@
...
+            +[...a.children][idx].textContent.replace(/[^0-9.]/g, '') <
+            +[...b.children][idx].textContent.replace(/[^0-9.]/g, '')
+          ) {
+            return -1;
+          }
+
+          return 0;
+        });
+
+        elem.classList.remove('DESC');
```
TarasHoliuk, line 57: Using class to store this data is not the best approach. You can add some `data-order` attribute, or store the last clicked index in the local variable in the code. Then you can compare last clicked index with current clicked index and define order

src/scripts/main.js
```diff
@@ -1,3 +1,422 @@
...
+        sortedRows = [...table.tBodies[0].children].sort((a, b) => {
+          if (
+            +[...a.children][idx].textContent.replace(/[^0-9.]/g, '') >
+            +[...b.children][idx].textContent.replace(/[^0-9.]/g, '')
+          ) {
+            return 1;
+          }
+
+          if (
+            +[...a.children][idx].textContent.replace(/[^0-9.]/g, '') <
```
TarasHoliuk, line 48: you can create a function to normalize it instead of repeating code a few times

src/scripts/main.js
```diff
@@ -1,3 +1,422 @@
...
+  button.setAttribute('type', 'submit');
+  button.textContent = 'Save to table';
+
+  return button;
+}
+
+function createSelect() {
+  const select = document.createElement('select');
+
+  const cities = [...table.tBodies[0].children]
```
TarasHoliuk, line 172: It's better to have defined constant list of cities

src/scripts/main.js
```diff
@@ -1,3 +1,422 @@
...
+    });
+
+    previousValue = cellValue;
+    inputBlur(input, cellValue);
+  });
+}
+
+removeCellText();
+
+function inputBlur(element, value) {
```
TarasHoliuk, line 379: it seems it doesn't work

TarasHoliuk (changes requested):
Hi! It looks good, but editing is broken

Also, I see tests cannot start at all now. If you couldn't solve the problem in the chat with a mentor but tried it - add the corresponding comment (with the link on thread in the chat would be perfect)

src/scripts/main.js
Voloshynd: Repaired

src/scripts/main.js
Voloshynd: constant list of cities is defined

src/scripts/main.js
Voloshynd: normalize function is created

src/scripts/main.js
Voloshynd: data-order attribute is added

PR #558, opened 2024-07-11
solution with optional

syavaYki:
[DEMO LINK](https://syavaYki.github.io/js_employees_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,222 @@
 'use strict';
 
-// write code here
+function sortRows(rows, sortIndex, accending) {
+  rows.sort((a, b) => {
+    const aClean = a.cells[sortIndex].textContent
+      ?.replace('$', '')
+      .replace(',', '');
+    const bClean = b.cells[sortIndex].textContent
+      .replace('$', '')
+      .replace(',', '');
+
+    if (isNaN(aClean)) {
+      if (accending) {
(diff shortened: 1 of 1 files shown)


anastasiiavorobiova (changes requested):
Great start! The editing works just for the one field however editing the second field for the same user doesn't. Also, try to fix tests

src/scripts/main.js
```diff
@@ -1,3 +1,234 @@
...
+const formElem = document.body.querySelector('#form-new-empl');
+
+for (let i = 0; i < tHead?.rows[0].cells.length; i++) {
+  sortMapIndx[headerCells[i].textContent] = i;
+}
+
+tHead?.addEventListener('click', (e) => {
+  const acdStat = e.target.classList.contains(className);
+
+  sortRows(tableRows, sortMapIndx[e.target.textContent], !acdStat);
```
TarasHoliuk, line 155: It seems like the newly added rows are not taken into account during sorting (always on the top)
You can query all rows again inside this function to make sure you use all rows

TarasHoliuk (changes requested):
Great job! One more fix required

anastasiia-tilikina (changes requested):
Almost done!
one more thing

Editing fields are not validated I can write invalid data. Let's fix it
image

vadiimvooo (approved):
LGTM

PR #560, opened 2024-07-15
js employees table dom

vlad-fil:
[DEMO LINK](https://vlad-fil.github.io/js_employees_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,214 @@
 'use strict';
 
-// write code here
+const thead = document.querySelector('thead');
+const tbody = document.querySelector('tbody');
+
+// table sorting by clicking on the title (in two directions)
+const nodeList = [...tbody.children].map((i) => i.innerText.split('\t'));
+let sortByName = 'asc';
+let sortByPosition = 'asc';
+let sortByOffice = 'asc';
+let sortByAge = 'asc';
+let sortBySalary = 'asc';
+
+thead.onclick = (e) => {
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,214 @@
...
+const thead = document.querySelector('thead');
+const tbody = document.querySelector('tbody');
+
+// table sorting by clicking on the title (in two directions)
+const nodeList = [...tbody.children].map((i) => i.innerText.split('\t'));
+let sortByName = 'asc';
+let sortByPosition = 'asc';
+let sortByOffice = 'asc';
+let sortByAge = 'asc';
+let sortBySalary = 'asc';
```
TarasHoliuk, line 12: you can store only the last title/th index that was clicked. Then compare current clicked title with last clicked - if it's the same order is 'desc'

src/scripts/main.js
```diff
@@ -1,3 +1,214 @@
...
+
+thead.onclick = (e) => {
+  const sortBy = e.target.innerText;
+
+  if (sortBy === 'Name') {
```
TarasHoliuk, line 17: use switch construction here

TarasHoliuk: Also, you can join some cases - the sorting logic is the same in a few different columns

Example (schematic):

```
case 'Name':
case 'Position':
  list.sort()
```

VitaliyBondarenko1982 (approved):
Looks good.

PR #562, opened 2024-07-16
Solution

Galers:
Have a nice day!😊😊😊
[DEMO LINK](https://Galers.github.io/js_employees_table_DOM/)


anastasiia-tilikina (changes requested):
Good start!
There are bugs we need to fix

1. I can add an employee with a negative number in the salary input
2. after adding an employee to the table it will disappear on sort click
3. There is no validation for fields on editing so I can pass invalid data (e.g. string in salary column)
4. 
image

Also we have tests failed, but lets start form these explicit errors and move on)

volodymyr-soltys97 (approved):
Great work 🔥

PR #567, opened 2024-07-18
Develop

siefimov:
[DEMO LINK](https://siefimov.github.io/js_employees_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -119,6 +119,8 @@ document.querySelectorAll('thead th').forEach((header) => {
     tableBody.innerHTML = '';
 
     insertSortedRows(sortedRows);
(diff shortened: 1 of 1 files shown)


anastasiia-tilikina (changes requested):
Very well done!
Good structured code!

There are some things we need to fix

image

1. I shouldn't be able to add negative number for salary
2. When I edit cell I shouldn't be able to add invalid data

volodymyr-soltys97 (changes requested):
Almost done!
Let's improve your code
The name and position cannot be only spaces, you need to fix it 
image

volodymyr-soltys97 (approved):
Nice work! 🔥

PR #573, opened 2024-07-22
add task solution

MariiaSuper:
https://MariiaSuper.github.io/js_employees_table_DOM/

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 1. Replace `<your_account>` with your Github username in the link
-    - [DEMO LINK](https://<your_account>.github.io/js_employees_table_DOM/)
+    - [DEMO LINK](https://MariiaSuper.github.io/js_employees_table_DOM/)
 2. Follow [this instructions](https://mate-academy.github.io/layout_task-guideline/)
     - Run `npm run test` command to test your code;
     - Run `npm run test:only -- -n` to run fast test ignoring linter;
--- a/package-lock.json
+++ b/package-lock.json
@@ -141,9 +141,9 @@
       }
     },
     "@babel/eslint-parser": {
-      "version": "7.23.10",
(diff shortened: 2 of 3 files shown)


SanyaBratashchuk (changes requested):
image
Validation should be applied on editing as well 
image
Notification doesn't disappear

volodymyr-soltys97 (changes requested):
Hi, the demo link still not fixed from the previous review, you need to make a deploy after your changes and check it after on the demo. 
Empty message when we don't add salary 
image
image

It will be better to delete this block from the message 3 seconds after it appeared

MariiaSuper:
> Hi, the demo link still not fixed from the previous review, you need to make a deploy after your changes and check it after on the demo. Empty message when we don't add salary image image
> 
> It will be better to delete this block from the message 3 seconds after it appeared

Sorry, did

MariiaSuper:
> > Hi, the demo link still not fixed from the previous review, you need to make a deploy after your changes and check it after on the demo. Empty message when we don't add salary image image
> > It will be better to delete this block from the message 3 seconds after it appeared
> 
> Sorry, my block disappears after 3 seconds. please explain what is the problem? thank you

volodymyr-soltys97 (changes requested):
Hi again, the demo link still not fixed, you need to run deploy again `npm run deploy`.
if you have problems with deploy feel free in the fe_chat for the help

MariiaSuper:
> Hi again, the demo link still not fixed, you need to run deploy again `npm run deploy`. if you have problems with deploy feel free in the fe_chat for the help

I did it couple of times, but I still do not see my changes was deployed... (((

image

anastasiiavorobiova (approved):
Great work!

PR #575, opened 2024-07-22
add task solution

ivankovbohdan:
[DEMO LINK](https://ivankovbohdan.github.io/js_employees_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,320 @@
 'use strict';
 
-// write code here
+const body = document.querySelector('body');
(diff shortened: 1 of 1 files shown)


etojeDenys (changes requested):
1. I should be able to sort array with my rows
https://github.com/user-attachments/assets/a698159a-e868-4c48-b820-18e4fb5af89b

etojeDenys (changes requested):
another issue, why I cannot edit created row
image

volodymyr-soltys97 (approved):
Good job 👍

PR #584, opened 2024-07-30
js_employees_table_DOM

svyatoslavPy:
[DEMO LINK](https://svyatoslavPy.github.io/js_employees_table_DOM/) 

description about tests :)

all tests passed , but one test has problem with name when we write our name in input and show message if we have less than 4 word in texts, i think this some wrong in tests. 

Screenshot 2024-07-30 at 15 15 59

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,311 @@
 'use strict';
 
 // write code here
+
+const body = document.querySelector('body');
+const table = document.querySelector('table');
+const tableHead = table.querySelector('thead');
+
+const tableContent = table.querySelector('tbody');
+const tableList = [...table.querySelectorAll('tbody tr')];
+
+const SORT_ORDERS = {
+  asc: 'asc',
+  desc: 'desc',
+};
+
+const SORT_TYPES = {
+  name: 'name',
+  position: 'position',
+  office: 'office',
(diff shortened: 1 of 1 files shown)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
1. You need to hide the success message after adding the person 
image

2. The name and position cannot be only spaces 
image

3. There should be a comma after the thousand 
image

4. You need to fix the failed test, if you need help feel free in the fe_chat

etojeDenys (changes requested):
demo link doesn't work, try to fix that

SanyaBratashchuk (changes requested):
1. salary can't be negative 
image

svyatoslavPy:
i fixed this problem with input and i did minimal number for only positive numbers when you select, but if you look at site , maybe it  is not changed, because i think some problem with git deploy

volodymyr-soltys97 (changes requested):
GJ!
To improve:
1. This comment still not fixed from the previous review 
image

2. You need to hide this message after 3 seconds
image

3. Fix the NaN value after clicking sorting by salary 
image

anastasiiavorobiova (approved):
Great work!

PR #597, opened 2024-08-08
add task solution

vneholiuk:
[DEMO LINK](https://vneholiuk.github.io/js_employees_table_DOM/)


anastasiia-tilikina (changes requested):
GJ
We need to fix some issues here

1. something i wrong with error messages
image

2. I shouldn't be able to add negative numbers
image

3. this requirement is not implemented
image

anastasiiavorobiova (changes requested):
Make sure to deploy your solution

PR #607, opened 2024-08-19
Develop

VikaChereushenko:
[DEMO LINK](https://VikaChereushenko.github.io/js_employees_table_DOM/)

I have one test failing. However, I successfully add a new employee to the table and Volodymyr Soltys has also checked it and proved it works fine. So I am a bit confused about it.

image


volodymyr-soltys97 (approved):
Great!

PR #610, opened 2024-08-20
js employees table DOM

Denys209818:
[DEMO LINK](https://denys209818.github.io/js_employees_table_DOM)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -141,9 +141,9 @@
       }
     },
     "@babel/eslint-parser": {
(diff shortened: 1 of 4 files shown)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
image
image

etojeDenys (changes requested):
better to add the logic to validate editing too
image

volodymyr-soltys97 (approved):
Great! 🔥

PR #619, opened 2024-08-26
js_employees_table_DOM

Valentin-1988:
[DEMO LINK](https://Valentin-1988.github.io/js_employees_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,207 @@
 'use strict';
 
 // write code here
+const headers = document.querySelectorAll('th');
+let rows = [...document.querySelectorAll('tbody tr')];
+const body = document.querySelector('tbody');
+const sortOrder = Array(headers.length).fill(true);
+
+function updateRows() {
+  rows = [...document.querySelectorAll('tbody tr')];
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,207 @@
...
+  sortOrder[index] = !sortOrder[index];
+}
+
+headers.forEach((head, index) => {
+  head.addEventListener('click', () => {
+    sort(index);
+  });
+});
+
+rows.forEach((row) => {
```
TarasHoliuk, line 45: Instead of adding a listener for each row add it to the table body and use the event bubbling concept to do something with the row. [Event delegation](https://www.freecodecamp.org/news/event-delegation-javascript/)

Would be nice to do similar with headers

TarasHoliuk (changes requested):
Good job! Let's add 1 improvement in your code

Valentin-1988:
[DEMO LINK](https://Valentin-1988.github.io/js_employees_table_DOM/)

etojeDenys (approved):
lgtm

PR #632, opened 2024-09-08
solution

oskushnir:
[DEMO LINK](https://oskushnir.github.io/js_employees_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,402 @@
 'use strict';
 
-// write code here
+const body = document.querySelector('body');
+const table = body.querySelector('table');
+
+const thead = table.querySelector('thead');
+const trThead = thead.querySelector('tr');
+const thThead = trThead.querySelectorAll('th');
+
+const tbody = table.querySelector('tbody');
+const trTbody = tbody.querySelectorAll('tr');
+
+const arrayOfPeople = [];
+const people = [];
+
+let isAscending = true;
+let activePerson = null;
+let clickTimeout = null;
+
+trTbody.forEach((tr) => {
+  const tds = tr.querySelectorAll('td');
+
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,402 @@
...
+                person.office = newValue;
+                break;
+
+              case 3:
+                person.age = +newValue;
+                break;
+
+              case 4:
+                person.salary = parseInt(newValue.replace(/[$,]/g, ''));
+                break;
```
vadiimvooo, line 196: You have to move this logic to some function according to dry principle.

src/scripts/main.js
```diff
@@ -1,3 +1,402 @@
...
+    );
+  } else if (positionValue.length === 0) {
+    pushNotification(
+      10,
+      10,
+      'Position error',
+      'You enter wrong position',
+      'error',
+    );
+  } else if (ageValue < 18) {
```
vadiimvooo, line 364:
```suggestion
  if (nameValue.length < 4) {
    pushNotification(
      10,
      10,
      'Name error',
      'Name should be more than 3 letters',
      'error',
    );
    
    return;
  } 
  
   if (positionValue.length === 0) {
    pushNotification(
      10,
      10,
      'Position error',
      'You enter wrong position',
      'error',
    );
    
    return;
  }
   if (ageValue < 18) {
```

vadiimvooo: Simplify the logic with conditions.

vadiimvooo (changes requested):
Good job, but you have to pass all tests. Feel free to ask for help in fe_chat.

volodymyr-soltys97 (approved):
Great work! 🔥

PR #635, opened 2024-09-11
add task solution

StanislavKapytsia:
- [DEMO LINK](https://StanislavKapytsia.github.io/js_employees_table_DOM/)

1. Implement table sorting by clicking on the title (in two directions).
- When users clicks on one of the table headers, table should be sorted in ASC order, the second click sorts it in DESC order.
- When users click on a new title, always sort in ASC order.

2. When user clicks on a row, it should become selected.
- Use 'active' class for table row to indicate it is selected.
- Only one line can be selected at a time.

3. Write a script to add a form to the document. Form allows users to add new employees to the spreadsheet.

4. Show notification if form data is invalid (use notification from the previous tasks).
- Click on `Save to table` should run validation for form inputs. If data is valid, add a new employee to the table.
- If `Name` value has less than 4 letters, show error notification.
- If `Age` value is less than 18 or more than 90 show error notification.
- If a new employee is successfully added to the table, show success notification.
- Notification titles and descriptions are up to you.
- Add qa attribute for notification: `data-qa="notification"` and class `error`/`success` depending on the result.

5. Implement editing of table cells by double-clicking on it. (optional)
- Double click on the cell of the table, should remove text, and append input with `cell-input` class.
- Only one cell can be edited at the time.
- On blur save changes to table cell. Remove input and set new text.
- On 'Enter' keypress, save changes to the table cell. Remove input and set new text in the table cell.
- If an input is empty on submitting return initial value.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 1. Replace `<your_account>` with your Github username in the link
-    - [DEMO LINK](https://<your_account>.github.io/js_employees_table_DOM/)
+    - [DEMO LINK](https://StanislavKapytsia.github.io/js_employees_table_DOM/)
 2. Follow [this instructions](https://mate-academy.github.io/layout_task-guideline/)
     - Run `npm run test` command to test your code;
     - Run `npm run test:only -- -n` to run fast test ignoring linter;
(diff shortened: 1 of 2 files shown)


VitaliyBondarenko1982 (approved):
Looks good

PR #638, opened 2024-09-14
task is done

BurchakDmitry:
[DEMO LINK](https://BurchakDmitry.github.io/js_employees_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,259 @@
 'use strict';
+// #region table parameters
 
-// write code here
+const formAtributes = {
+  name: 'Name',
+  position: 'Position',
+  office: 'Office',
+  age: 'Age',
+  salary: 'Salary',
+};
+
+const citiesSelect = [
+  'Tokyo',
+  'Singapore',
+  'London',
+  'New York',
+  'Edinburgh',
+  'San Francisco',
+];
+
+// #endregion
+const table = document.querySelector('table');
+const thead = table.querySelector('thead');
+const cellsOfHead = [...thead.querySelectorAll('th')];
+const tbody = document.querySelector('tbody');
+const employee = newForm(formAtributes, citiesSelect);
+let clicksOnHeadCells = 1;
+let indexOfHeadCells = null;
+let indexOfRow = null;
+
+employee.classList.add('new-employee-form');
+document.body.appendChild(employee);
+
+// #region creating a notification
+function notificationMessage(text, type) {
+  const notification = document.createElement('div');
+  const titleNotification = document.createElement('h3');
+
+  notification.setAttribute('data-qa', 'notification');
+  notification.classList.add('notification');
+  titleNotification.classList.add('title');
+
+  notification.appendChild(titleNotification);
+
+  notification.classList.add(type);
+  titleNotification.textContent = text;
+
+  document.body.appendChild(notification);
+
+  setTimeout(() => {
+    document.body.querySelector('.notification').remove();
+  }, 10000);
+}
+// #endregion
+
+// #region salary formatting
+function formatCurrency(numberStr) {
+  const number = parseInt(numberStr, 10);
+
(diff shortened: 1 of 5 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,259 @@
...
+
+    if (parseInt(inputAge.value) < 18) {
+      notificationMessage('Your age is too young :(', 'error');
+    }
+
+    if (parseInt(inputAge.value) > 90) {
+      notificationMessage('Your age is too old :(', 'error');
+    }
+
+    if (inputPosition.value.trim() === '') {
```
TarasHoliuk, line 172:
```suggestion
    if (!inputPosition.value.trim()) {
```
shorter version

But also, you should show the error if age and salary inputs are empty. Now when I click Save button without these inputs nothing happens:

image

src/scripts/main.js
```diff
@@ -1,3 +1,259 @@
...
+  ];
+
+  if (
+    inputName.value.trim() !== '' &&
+    inputName.value.length >= 4 &&
+    inputAge.value.trim() !== '' &&
+    parseInt(inputAge.value) >= 18 &&
+    parseInt(inputAge.value) <= 90 &&
+    inputPosition.value !== '' &&
+    inputSalary.value !== ''
```
TarasHoliuk, line 138: So now you check it twice - here, inside the if condition, and below to show the error message. Instead, validate all values in a separate function. If some value is invalid show error message (only one at the time to prevent overlapping).

Return true/false from the function and use this value here

src/scripts/main.js
```diff
@@ -1,3 +1,259 @@
...
+
+// #region form submit event
+employee.querySelector('button').addEventListener('click', (e) => {
+  const inputName = employee.querySelector('[data-qa="name"]');
+  const inputPosition = employee.querySelector('[data-qa="position"]');
+  const inputOffice = employee.querySelector('[data-qa="office"]');
+  const inputAge = employee.querySelector('[data-qa="age"]');
+  const inputSalary = employee.querySelector('[data-qa="salary"]');
+
+  const valuesOfInputs = [
```
TarasHoliuk, line 123: it's input elements yet, not their values

src/scripts/main.js
```diff
@@ -1,3 +1,259 @@
...
+    if (inputName.value.length < 4) {
+      notificationMessage('Name length is less than 4 characters', 'error');
+    }
+
+    if (parseInt(inputAge.value) < 18) {
+      notificationMessage('Your age is too young :(', 'error');
+    }
+
+    if (parseInt(inputAge.value) > 90) {
+      notificationMessage('Your age is too old :(', 'error');
```
TarasHoliuk, line 169:
```suggestion
      notificationMessage('Allowed age is between 18 and 90', 'error');
```
"age is too old" is a wrong sentence. You can use suggested for both cases (< 18, > 90) - it also says user which value is allowed

src/scripts/main.js
```diff
@@ -1,3 +1,259 @@
...
+  }
+
+  return rows;
+}
+
+// #endregion
+
+// #region formatting a string to a number
+function stringToNumber(str) {
+  if (str.includes(',')) {
```
TarasHoliuk, line 205: Instead of checking it, you can just remove all `','` in any case. Also, now sort doesn't work for salary (you forget about $ sign)

src/scripts/main.js
```diff
@@ -1,3 +1,259 @@
...
+    }
+    indexOfRow = index;
+
+    row.classList.add('active');
+  });
+});
+// #endregion
+
+// #region sorting in two directions
+cellsOfHead.forEach((cell, index) => {
```
TarasHoliuk, line 234: use [event delegation](https://www.freecodecamp.org/news/event-delegation-javascript/) instead of adding event to each header separately using loop

TarasHoliuk (changes requested):
GJ 👍  But some fixes are required

PR #639, opened 2024-09-14
first draft, but all basic tasks i done

MaxF1996:
- [DEMO LINK](https://MaxF1996.github.io/js_employees_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,238 @@
 'use strict';
 
 // write code here
+
+const pageTable = document.querySelector('table');
+
+const pageTableState = {};
+
+function extractAmount(str) {
+  return parseFloat(str.replace(/[^0-9.]/g, ''));
+}
+
+function convertToCurrency(val) {
+  return val.toLocaleString('en-US', {
+    style: 'currency',
+    currency: 'USD',
+    minimumFractionDigits: 0,
+    maximumFractionDigits: 0,
+  });
+}
+
+function validateEmployee(employee) {
+  const { employeeName, position, office, age, salary } = employee;
+
+  if (
+    !employeeName ||
+    employeeName.length < 4 ||
+    !position ||
+    !office ||
+    +age < 18 ||
+    +age > 90 ||
+    !salary ||
+    Number.isNaN(+salary)
+  ) {
+    showNotification('warning');
+
+    return false;
+  } else {
+    showNotification('success');
+
+    return true;
+  }
+}
+
+function showNotification(type) {
+  const notification = document.createElement('div');
+  const title = document.createElement('p');
+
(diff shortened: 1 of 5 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,279 @@
...
+    const select = document.createElement('select');
+
+    if (types[i] !== 'select') {
+      label.append(input);
+      input.setAttribute('data-qa', inputs[i]);
+      input.setAttribute('type', types[i]);
+
+      if (inputs[i] === 'age') {
+        input.setAttribute('min', '18');
+        input.setAttribute('max', '120');
```
TarasHoliuk, line 108: should be 90

src/scripts/main.js
```diff
@@ -1,3 +1,279 @@
...
+function newEmployeeForm() {
+  const employeeForm = document.createElement('form');
+
+  employeeForm.classList.add('new-employee-form');
+  employeeForm.setAttribute('novalidate', '');
+  employeeForm.setAttribute('id', 'new-employee-form');
+
+  const labels = ['Name', 'Position', 'Office', 'Age', 'Salary'];
+  const inputs = ['name', 'position', 'office', 'age', 'salary'];
+  const types = ['text', 'text', 'select', 'number', 'number'];
```
TarasHoliuk, line 83: instead of 3 separate arrays it's better to create one with objects like:
```
{
  name: 'name',
  label: 'Name',
  type: 'text',
}
```

src/scripts/main.js
```diff
@@ -1,3 +1,279 @@
...
+      row.classList.toggle('active', row === pageTableState.selectedRow);
+    });
+  }
+});
+
+function saveChanges(input, initialValue) {
+  const newValue = input.value.trim() || initialValue;
+  const cell = input.parentElement;
+
+  cell.textContent = newValue;
```
TarasHoliuk, line 244: for salary would be nice to transform it in the same way as you do when adding a new row

src/scripts/main.js
```diff
@@ -1,3 +1,279 @@
...
+
+  if (
+    !employeeName ||
+    employeeName.length < 4 ||
+    !position ||
+    !office ||
+    +age < 18 ||
+    +age > 90 ||
+    !salary ||
+    Number.isNaN(+salary)
```
TarasHoliuk, line 33: It would be nice to show a bit more informative errors in each case depending on which value is wrong, for example: `Age is required` or `Allowed age is between 18 and 90`, etc

TarasHoliuk (changes requested):
GJ 👍 
Let's add a few improvements

vadiimvooo (approved):
Well done

PR #641, opened 2024-09-17
my solution

yursprava:
[DEMO LINK](https://yursprava.github.io/js_employees_table_DOM/)

The code fails the tests that require adding classes for the 'notification' element. But, such classes are added correctly in my code. Please tell me what is the problem here?

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,374 @@
 'use strict';
 
-// write code here
+const tbody = document.querySelector('tbody');
+
+const thead = document.querySelector('thead');
+
+const thCollection = document.querySelectorAll('th');
+
+for (const currTh of thCollection) {
+  currTh.setAttribute('data-state', 'unsorted');
+}
+
+function ascSortStringlocaleCompare(index) {
+  const tbodyRows = document.querySelectorAll('tbody tr');
+  const rowsArray = Array.from(tbodyRows);
+
+  rowsArray.sort((a, b) => {
+    const ageA = a.cells[index].innerHTML;
+
+    const ageB = b.cells[index].innerHTML;
+
+    return ageA.localeCompare(ageB);
+  });
+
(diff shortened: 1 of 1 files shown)


volodymyr-soltys97 (changes requested):
GJ!
To improve:
The form fields cannot be only spaces, fix it 
image

Is you need help with failed tests feel free in the fe_chat

yursprava:
Solution was fixed

src/scripts/main.js
```diff
@@ -1,3 +1,428 @@
...
+button.setAttribute('type', 'submit');
+button.innerHTML = 'Save to table';
+button.classList.add('button');
+form.append(button);
+
+button.addEventListener('click', function (ev) {
+  ev.preventDefault();
+  notification.removeAttribute('style');
+
+  const err1 = nameInput.value.length < 4;
```
volodymyr-soltys97, line 318: Use a more descriptive name for error, fix it everywhere
```suggestion
  const nameLengthError = nameInput.value.length < 4;
```

src/scripts/main.js
```diff
@@ -1,3 +1,428 @@
...
+      ev.target.setAttribute('data-state', 'ascending');
+      ascSortSalary(index);
+    } else if (ev.target.dataset.state === 'ascending') {
+      ev.target.setAttribute('data-state', 'descending');
+      descSortSalary(index);
+    } else if (ev.target.dataset.state === 'descending') {
+      ev.target.setAttribute('data-state', 'ascending');
+      ascSortSalary(index);
+    }
+  } else {
```
volodymyr-soltys97, line 144: Consider will be better if you use switch case construction here and move some logic to the helper functions

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

yursprava:
fixed

volodymyr-soltys97 (approved):
Great work! 🔥

PR #647, opened 2024-09-19
first solution

Yexelay:
- [DEMO LINK](https://Yexelay.github.io/js_employees_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,272 @@
 'use strict';
 
-// write code here
+const table = document.querySelector('table');
+const headers = table.querySelectorAll('th');
+
+let sortDirection = 1;
+let currentIndex = -1;
+let activeInputCell = null;
+
+const removeActiveClass = () => {
+  table.querySelectorAll('tr').forEach((row) => row.classList.remove('active'));
+};
+
+const removeActiveInput = () => {
+  if (activeInputCell) {
+    const input = activeInputCell.querySelector('input');
+    const originalText = input.getAttribute('data-original-value');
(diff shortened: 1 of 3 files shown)


etojeDenys (changes requested):
looks good, but it's better to add the logic to validate the editing inputs
image

Yexelay:
If we work with the data that we added through the form, the validations pass correctly: the name can only contain letters, and the age must be numbers between 18 and 90. However, the data that was already in the table bypasses the validity checks. How can this be fixed?

Yexelay:
And also, for some reason, one of the tests didn't pass, even though the validation is being performed
изображение_2024-09-20_112604148
изображение_2024-09-20_112623846
изображение_2024-09-20_112705423

volodymyr-soltys97 (changes requested):
Good job 👍 
image

Also, if you need help with tests feel free in the fe_chat

src/scripts/main.js
```diff
@@ -1,3 +1,308 @@
...
+
+    return aText.localeCompare(bText) * direction;
+  });
+
+  tableBody.innerHTML = '';
+  tableBody.append(...sortedRows);
+};
+
+const form = document.createElement('form');
+
```
anastasiiavorobiova, line 144:
```suggestion

form.setAttribute('novalidate', true);
```

anastasiiavorobiova (changes requested):
Great work! Added a comment to pass test

anastasiiavorobiova (approved):
Great work!

PR #662, opened 2024-10-07
add task solution

Nadinnnka:
- [DEMO LINK](https://Nadinnnka.github.io/js_employees_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,233 @@
 'use strict';
 
-// write code here
+const table = document.querySelector('table');
+const headers = document.querySelectorAll('th');
+const rows = Array.from(table.querySelectorAll('tbody tr'));
+const tbody = table.querySelector('tbody');
+
+let isAscending = true;
+let currentSortIndex = -1;
+
+const sortTable = (index) => {
+  rows.sort((rowA, rowB) => {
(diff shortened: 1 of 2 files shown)


anastasiiavorobiova (changes requested):
Great work!
Screenshot 2024-10-07 at 14 21 43
A user shouldn't be able to add NaN values

src/scripts/main.js
```diff
@@ -1,3 +1,234 @@
...
+
+    selectedRow.classList.add('active');
+  }
+});
+
+const dataAttributes = ['name', 'position', 'office', 'age', 'salary'];
+const form = document.createElement('form');
+const button = document.createElement('button');
+
+form.classList.add('new-employee-form');
```
anastasiiavorobiova, line 67: Consider adding `novalidate` attribute to the form

anastasiiavorobiova (changes requested):
Great work! The previous comment is not fixed

src/scripts/main.js
Nadinnnka: but I corrected the previous comment, if you think it's better to add the novalidate attribute to the form, I can change it

Знімок екрана 2024-10-07 144226

Знімок екрана 2024-10-07 144149

volodymyr-soltys97 (approved):
Good job! 👍

PR #717, opened 2024-11-25
Task solution

Dominik-Dudek-code:
- [DEMO LINK](https://Dominik-Dudek-code.github.io/js_employees_table_DOM/)

Hello,
What comes to my mind, is that - my code may seem quite long, but since this is the final task related to advanced JavaScript - DOM lessons, I have implemented a few things that I thought were necessary and would be useful as extras:

1. The entire script is based on the thead cells, which I don't reference directly but generate dynamically, from which I create an object that the rest of the script is based on. So adding additional thead category, automatically adding form input and sorting option.
2. Sorting is done using the dynamically generated thead cells, which sort according to the task requirements, new sorting option is set by default on strings.
3. I also created a form that dynamically generates inputs, created in accordance with the order of thead cells. If no parameters are passed, the generated input defaults to a text type and is automatically validated by the content within it.
4. Notifications were also handled a bit in my own way, because there isn't a specific class for that. I implemented an option where only one message is displayed at a time on the screen, preventing them from overlapping. Therefore, only one message, in the order of validation, is shown, disappearing after 5 seconds.
5. Additionally, the notifications have messages to indicate which fields are incomplete. I implemented a function from another task that lists the missing fields, using commas and "and" to separate them.
6. The last implementation I added is that when the table is currently sorted, any newly added item will be inserted according to the current sorting mode. If no sorting mode is set, the item will be added automatically at the end.

I’m happy to receive any feedback that comes to your minds!

Additionally, based on your feedback, I will proceed to implement optional features related to this task, such as table editing, and I also plan to add element numbering and while adding new category, creating empty cells, that can be edited and last but not least, scrolling when there are enough elements in the table.

Let me know what do you think about it, thanks!

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,345 @@
 'use strict';
 
-// write code here
+// select all table pieces
+const table = document.querySelector('table');
+const rows = document.querySelector('tbody');
+const head = document.querySelector('thead');
+
+// prevent selecting for headers
+head.style.userSelect = 'none';
+
+// sorting parameters
+const sortingParameters = {
+  lastClickedCell: null,
+  sortOrder: 'asc',
+
+  get sortingName() {
+    return this.lastClickedCell;
+  },
+
+  set sortingName(value) {
+    return (this.lastClickedCell = value.textContent.toLowerCase());
+  },
+};
+
+// column types
+const columnIndexMap = {};
+
+head.querySelectorAll('th').forEach((cell, index) => {
+  const columnName = cell.textContent.trim().toLowerCase();
+
+  columnIndexMap[columnName] = index;
+});
+
+// event for sorting
+table.addEventListener('click', (e) => {
+  if (e.target.nodeName !== 'TH' || e.target.closest('thead') === null) {
+    return;
+  }
+
+  const sortBy = e.target.textContent.toLowerCase();
+
+  if (sortBy === sortingParameters.lastClickedCell) {
+    sortingParameters.sortOrder =
+      sortingParameters.sortOrder === 'asc' ? 'desc' : 'asc';
+  }
+
+  sortingTable(rows, sortBy, sortingParameters.sortOrder);
+
+  sortingParameters.sortingName = e.target;
+});
+
+// function for sorting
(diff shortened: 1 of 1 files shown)


danon321 (approved):
Wyglada i działa bardzo dobrze! Dobra robota :)))

Jedyne co mogłbym zasugerować to rozbić to na pomniejsze pliki/skrypty. Tak aby kazdy z nich zajmował sie swoja działką. Jeden np tworzył formularz a drugi zajmował sie dodawaniem uczestników

Dominik-Dudek-code:
Hello again,

as I mentioned earlier, I’ve implemented new functionalities into the table that expand the project with several new features.

ID Column
I added an ID column, which is automatically assigned to each user in the table. Each user has a unique ID that cannot be edited in any way, including via double-click. This field is locked and remains unmodifiable.

Delete Icon with Confirmation Modal
A new feature is the delete icon that animates on the right side of the selected row. Clicking the icon triggers a modal to confirm whether the user really wants to delete the row. For this, I created a dedicated structure in the HTML to avoid handling this entirely in JavaScript.

Improved Notification Display
I enhanced the way notifications are displayed. Notifications now validate the form input properly and provide clear feedback about validation status.

Editable Cells with Validation
When double-clicking a table cell, its content becomes editable. The input is validated in the same way as the form fields. On blur or pressing Enter, the cell’s value is validated, and a notification appears only if the input is invalid. Additionally, text inputs automatically capitalize the first letter of each word.

Dynamic Form and Sorting Integration
To streamline the process, I implemented an object that maps the form fields and the table headers. This ensures that any new column automatically has sorting functionality and generates a corresponding input field in the form. By default, the new input fields have no assigned validation mode until manually configured.

At this point, I plan to leave the table as is for the next few days to continue progressing with the course. However, I’ll revisit the table before the holidays to refine it further.

This brings me to a couple of questions:

What else could be improved or added as a new feature? The only idea I have so far is adding a feature below the form that allows the creation of new columns, with options for automatic filling and editing.
Is the code well-structured and efficient? Are there any areas that could be simplified or optimized further?
I would love to hear your thoughts and suggestions!

PR #814, opened 2025-03-04
solution

yuliia-nudyk:
DEMO - https://yuliia-nudyk.github.io/js_employees_table_DOM/


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
The form fields cannot be only spaces 
image
image

PR #856, opened 2025-04-15
add task solution

hannachachacha:
- [DEMO LINK](https://hannachachacha.github.io/js_employees_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,234 @@
 'use strict';
 
-// write code here
+const tableHeaders = document.querySelectorAll('th');
+const tableBodyRows = document.querySelectorAll('tbody tr');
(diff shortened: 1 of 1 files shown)


hannachachacha:
@mateacademy-ai-mentor 
1) "It's good to declare variables with const or let instead of var to avoid hoisting issues and ensure better code scoping." I did not use var, I do not understand what this comment is about
2) “You should ensure that only the clicked row has the active class.” - yes, that's why I remove this class for all other rows first: r.classList.remove('active');
3) "The isNaN(Number(salaryValue)) check only verifies that the salary value can be converted to a number, but it doesn't check if the salaryValue is empty." This is a field with the `required` attribute, it cannot be empty

PR #903, opened 2025-06-13
add task solution

gtailde:
- [DEMO LINK](https://gtailde.github.io/js_employees_table_DOM/)


gtailde:
In my opinion, there is a bug, if I just edit the cell, everything is ok, but if I try to select a word, delete it and save the cell, then for some reason it is saved empty, I have already tried different methods but nothing helped, I wonder why this happens, if in the console I get the correct values ​​that should be set.

VitaliyBondarenko1982 (approved):
Good job.

